PR #1, opened 2022-08-08
feat: add storybooking to stonecrop

Alchez:
**Commands**:
- `rushx story:dev` to start the histoire dev-server

**Changes:**
- Added histoire and vue-plugin libs to the atable package
- Moved the existing atable stories to histoire
- Updated vite and its vue plugin to latest on all packages
- Updated typings on vite, and imported typings for histoire components

<hr>

**Screenshots:**

image

image


agritheory:
@Alchez Have a look at the readme for development workflow using Rush and pnpm. At no time should `npx` be invoked - use `rushx` instead. Stories should be running under the `rushx dev` command , per library. (Alternatively, it may make sense to have a `story` in `scripts` instead of a `dev` command).

agritheory:
@Alchez RE committing to dist, I think it's correct to do so. [Vue does it](https://github.com/vuejs/vue/tree/main/dist)

Alchez:
> @Alchez Have a look at the readme for development workflow using Rush and pnpm. At no time should `npx` be invoked - use `rushx` instead.

Oh, that's good to know. I think it'll take some time for me to get used to these tools too.

> Stories should be running under the `rushx dev` command , per library. (Alternatively, it may make sense to have a `story` in `scripts` instead of a `dev` command).

That makes sense. I was planning to put it into a script eventually. I'll add a new `story` script.

> @Alchez RE committing to dist, I think it's correct to do so. [Vue does it](https://github.com/vuejs/vue/tree/main/dist?rgh-link-date=2022-08-08T11%3A55%3A01Z)

Alright, I'll keep the changes in then.

agritheory:
After messing around with dependencies yesterday, I think we're better off without a Nuxt dependency in this package, but it probably makes sense as part of the examples package.  The Nuxt stories package has a ton of dependencies and is based on Nuxt 2 / webpack and while I like the idea, those are a deal breaker. Let's cancel the Nuxt todo and skip to migrating the two stories over.

agritheory (changes requested):
This looks great. Can we add stub for the documentation field and one for events (if applicable)? 
image

Just saw your edit to the first message with this question. Yes, docs please.

agritheory:
We are also going to want to theme this to be in line with brand/framework themes but we can push that off until we have a larger design for the whole thing - this should be before publishing it. https://histoire.dev/guide/config.html#theming

PR #8, opened 2022-08-12
[project] enforce prettier in CI

agritheory:


agritheory:
This needed a baseline format of all files to the new prettier spec. Future work will fail if not formatted.

PR #10, opened 2022-08-17
[aform] setup masking for AForm components

Alchez:
**Ref:** https://erp.agritheory.com/app/task/TASK00131


agritheory:
@Alchez Is this ready? Do you want to add ADate to a story first?

Alchez:
@agritheory So ADate currently assumes that it lives inside a table always, and all its event handlers try to update the table data.

I think we might want to consider an ADate refactor that maybe includes an input field that it can target? But it should also be viable inside a table. I think we can discuss more today.

I've marked this PR as ready though, since input masking is available for some generic inputs as well as custom functions.

agritheory (approved):
Stories look good, lots of positive changes in here.

PR #11, opened 2022-08-24
[project] create rush-prettier autoinstaller

Alchez:


agritheory:
Rebased and included in #12

PR #14, opened 2022-08-25
[utilities] create composable for keyboard navigation

### LINKED ISSUE #15: [utilities] keyboard navigation for hidden elements
**Problem:** Handling keyboard navigation for hidden elements (e.g. collapsible tree-views)

**Context:** There is [no single perfect](https://stackoverflow.com/a/59096915) way to detect hidden elements in a DOM tree using Javascript

**Possible Solution(s):**

- We implement a combination of the techniques in the above link to filter out hidden elements in the composable, and not attach the listener to those.
  - **Pro:** This will not break keyboard navigation, since hidden elements are skipped and therefore focus events will not fail.
  - **Con:** This is implicit behaviour, since the hidden selectors will not receive the event listener. This can be made explicit with a key in the config object.
- We defer the responsibility to the user to filter out the hidden elements before passing it to the composable.
  - **Pro:** The composable is lighter and more reliable, without having to deal with another function and possible edge cases.
  - **Con:** If this expectation is not clear and the user passes hidden elements, that can break keyboard navigation, and that may be bad UX for someone using the composable. This can be avoided with API documentation and/or a warning/error/linter support.

### LINKED ISSUE #16: [utilities] applying event listeners to a limited set of selectors
**Problem:** Applying event listeners to a limited set of selectors (e.g. all cells in a table except the number column) in the context of keyboard navigation

**Context:** If a user provides a string as the selector input to the composable, we currently run `document.querySelectorAll` to fetch all matching nodes, which returns ALL nodes on that page, which may or may not be implicit.

**Possible Solution(s):**
- We keep the status quo.
  - **Pro:** The user has the responsibility to explicitly state what selectors they would like to apply the listeners on.
  - **Con:** This limits the use of the composable, since the user has to do more to have control over the set of selectors they want. This is fairly subjective, since JS can allow fine control in `querySelectorAll`.
- We accept a parent element and apply the `querySelectorAll` call on that instead
  - **Pro:** The user can easily manage scopes for the target elements, and re-use across different listener options.
  - **Con:** If this option is provided, we'll need to add a default to it. `document` sounds fairly reasonable. This can be made easier with API documentation.

**Additional Info:** Either way, we will also probably need to expand on the types of inputs that the `useKeyboardNav` composable should receive.

Alchez:
**Changes:**

- Added composable for keyboard navigation in tables
- Allow accepting chained keypresses (Shift + Tab, Ctrl + C) as separate listeners to the composable
- Skip un-focusable and hidden elements during keyboard navigation
- Complete checklist defined in [#16](https://github.com/agritheory/stonecrop/issues/16#issuecomment-1235468230)
- Complete first milestone as defined in #18
- Complete second milestone as defined in #18
- Set ~~ADate's~~ ATableModal (or package-level equivalent) default focus to input date, or if not available, current date


agritheory (approved):
Looks great, works as expected.
Docs to be covered in future PR
Cypress tests to be included in Future PR

PR #22, opened 2022-09-16
[project] ci: allow publishing monorepo apps

### LINKED ISSUE #17: Omit built (/dist) packages from version control, use GitHub packages
Committing built files feels like an extra and unnecessary step

https://github.com/features/packages

Alchez:
Closes #17.


rush.json
```diff
@@ -11,8 +11,8 @@
 	},
 	"gitPolicy": {},
 	"repository": {
-		"url": "https://github.com/sedum-io/",
-		"defaultBranch": "develop",
+		"url": "https://github.com/Alchez/stonecrop",
```
agritheory: Can you change this to AgriTheory?

Alchez: I actually want to run some publishing tests on my fork first, I'll change it back when this is ready for review

Alchez:
@agritheory 

Okay, so I was able to publish the apps to my [fork](https://github.com/Alchez/stonecrop). You should be seeing them on the right.

However, we'll have to make some structural changes to get this to work on the `agritheory` Github org.

Github's package registry expects its packages to either be named `package-name` or `@scope/package-name`, where `scope` is the name of the organization.

So we'll either have to name the packages like `@agritheory/aform`, or `@sedum-io/aform`. https://github.com/sedum seems to be taken, so we won't be able to use that name.

Once we finalize that, we can confirm the new development workflow and I'll then add the build steps in the release file for Github Actions.

agritheory:
@Alchez Let's go with AgriTheory. I'll get in touch with sedum and see if they are willing to part with the name.

PR #23, opened 2022-09-27
[project] setup packages as modules

Alchez:


agritheory:
@Alchez What's the difference between minify: false and sourcemap: true?

PR #24, opened 2022-09-28
[atable] add story for table-list component

Alchez:
[Screencast from 29-09-22 12:18:47 PM IST.webm](https://user-images.githubusercontent.com/13396535/192959399-4b75307d-674e-4859-bb8c-00ca168ae31b.webm)


agritheory:
@Alchez Testing locally, I can still focus the individual cells and the keynav doesn't seem to work (except for tab). Can you mark this as draft?

Alchez:
@agritheory, I added a fix that handles the click input on the cell and focuses the row instead. Performance isn't great (it's not terrible either, but there's a visible switch), but it's the only way I can see it without disabling pointer events for cells, without which we cannot focus on the rows. Do you have any suggestions?

agritheory:
@Alchez This was my first thought - it doesn't improve the problem:
ACell Line 64
```js
const handleInput = (event: MouseEvent) => {
	if (props.clickHandler && props.tabIndex < 0) {
		props.clickHandler(event)
		return
	}
```
@johnhkelley1 John is back next week, let's have him take a stab at it.

I think the fallback plan would be to animate the outline so it fades in if its going to be that slow. Probably just as much work to make that hack look good as it would be to fix the underlying (unknown) issue.

agritheory:
@Alchez @johnhkelley1 Testing this locally and getting a feel for it...
The left/ right arrow keys to traverse rows is weird enough that I don't like it... like a 2/5 stars feature, would be happy to have you instincts on it as well. 

I realize that the "correct" key nav for this has been moving target as we try to figure out what feels right. I think I've over-complicated this. And after trying out the work you guys have done so far I think we want to make the follow changes.

Use the default key nav spec, EXCEPT up and down arrow keys and their variants (`'keydown.alt.down'`) focus the next row up or down. Tab and left/right arrows traverses cellwise, ignoring rows. we should have started with this as a spec (my fault for giving poor guidance).

| Key(s)        | Function    | Notes  |
| :------------ | :--------   | :----- | 
| Enter         | `downRowl`    | todo   |
| Tab           | `nextCell`    |        |
| Shift + Enter | `upRow`       | todo   |
| Shift+ Tab    | `prevCell`    |        |
| &#8592;       | `prevCell`    |        |
| &#8593;       | `upRow`       | todo   |
| &#8594;       | `nextCell`    |        |
| &#8595;       | `downRow`     | todo   |
| Home          | `firstRow`    | todo   |
| End           | `lastRow`     | todo   |
| Ctrl + &#8592;| `firstCell`   |        |
| Ctrl + &#8594;| `lastCell`    |        |

To more easily implement this and make this design a bit more future proof I think it makes sense to refactor the `addNavigation` property to accept a `useKeyboardNav` config partial: 
- `false` removes the keynav from the cell(s)
- `true` uses the default implementation
- `{'keydown.up' : rowWiseHandler } would override the `keydown.up` keypress specifically in the handlers argument of the useKeyboardNav composable.

We probably want to make this a common interface on the ARow component as well. It could be handy on some components in AForm as well, but that's out of scope for this PR.

agritheory:
@Alchez @johnhkelley1 Nice work guys, merging all your work with Rohan's original PR.

PR #31, opened 2022-10-12
[stonecrop] apply SPA design for stonecrop

### LINKED ISSUE #32: Remove opinioned listview/record pattern from registry and router
Doctypes should be responsible for their own routing logic. This enables single page/ singleton Doctypes and those with potentially complex routing opinions like a tree or Gantt view. While the list/record pattern will still likely end up being the majority case, it won't be correct for all cases.

Alchez:
Closes #32


stonecrop/src/index.ts
```diff
@@ -10,6 +10,10 @@ const Stonecrop = {
 	install: (app: App, options: InstallOptions) => {
 		app.use(router)
 		app.provide('$registry', new Registry(options.router || router, options.schemaLoader))
+
+		for (const [tag, component] of Object.entries(options.components)) {
+			app.component(tag, component)
+		}
```
agritheory, line 16: Excellent idea

stonecrop/stories/components/Records.vue
```diff
@@ -1,11 +1,12 @@
 <template>
-	<ATable :key="records.rows" :columns="records.columns" :rows="records.rows" :config="records.config" />
+	<ATable :key="records.columns" :columns="records.columns" :rows="records.rows" :config="records.config" />
```
agritheory, line 2: Can you explain this switch?

PR #33, opened 2022-10-14
[stonecrop] process events and hooks for doctypes

### LINKED ISSUE #34: FSM and Hook Implementation
Every Doctype shall have a finite state machine. These describe the flow of states /statuses of data in its user-facing context. Nested machines (an Order having an externally derived "Paid" or Delivered" state) are fine, as are trivial machines (an Access Log Doctype with only "Created" and "Saved" states for example). These FSM will be enforced on both the server and in the UI so they need to be mutually compatible/ client-agnostic. The FSM will fire hooks - events stored in a map as part of the metadata - to create mutations to the document and/or side effects. At this time hooks can be plain functions but it is likely that we will expand them to include more metadata in the future (client/server-only flags, permissions, idempotency implementation, etc).

This POC contains several scopes of development: 
- [x] FSM integration
- [x] Hooks implementation

The following are moved to #35 and out of scope of this Issue:
- A [`Workflow` UI component](https://github.com/agritheory/stonecrop/issues/35) allows the user to create a document and interact with its state (part of Stonecrop)
- A subset of the `Workflow` component will be a button / button group design effort (port of AForm)

Alchez:
Closes #34.

### DIFF
--- a/stonecrop/src/doctype.ts
+++ b/stonecrop/src/doctype.ts
@@ -1,25 +1,28 @@
-import { ComponentOptions } from 'vue'
+import { Component } from 'vue'
 
-import { ImmutableRegistry } from 'types/index'
+import { ImmutableDoctype } from 'types/index'
 
 export default class Doctype {
 	doctype: string
-	schema: ImmutableRegistry['schema']
-	events: ImmutableRegistry['events']
-	hooks: ImmutableRegistry['hooks']
+	schema: ImmutableDoctype['schema']
+	events: ImmutableDoctype['events']
+	hooks: ImmutableDoctype['hooks']
+	// TODO: allow different components for different views; probably
+	// should be defined in the schema instead?
+	component?: Component
 
 	constructor(
 		doctype: string,
-		schema: ImmutableRegistry['schema'],
-		events: ImmutableRegistry['events'],
-		hooks: ImmutableRegistry['hooks'],
-		component: string | ComponentOptions = undefined,
-		recordsComponent: string | ComponentOptions = undefined
+		schema: ImmutableDoctype['schema'],
+		events: ImmutableDoctype['events'],
+		hooks: ImmutableDoctype['hooks'],
+		component?: Component
 	) {
 		this.doctype = doctype
 		this.schema = schema
 		this.events = events
 		this.hooks = hooks
+		this.component = component
 	}
 
 	get slug() {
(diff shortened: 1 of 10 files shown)


stonecrop/src/doctype.ts
```diff
@@ -1,17 +1,17 @@
...
-import { SchemaTypes } from '@agritheory/aform/types'
+import { Meta } from 'types/index'
 
 export default class Doctype {
 	doctype: string
-	schema: SchemaTypes[]
-	events: any // TODO: new Machine()
-	hooks: any
+	schema: Meta['schema']
+	events: Meta['events']
```
agritheory: I don't think this is right. These need to be immutable primitives. Meta as a type is duplication of the Doctype class / type.

Alchez: I can change the values to read-only, but did you want me to change the name as well?

```suggestion
	readonly events: Meta['events']
```

agritheory: I was thinking of something like this:

```
import { Map } from immutablejs
import { MachineConfig } from 'xstate'

schema:  Map
readonly events: MachineConfig
hooks: Map
```

`MachineConfig` isn't immutable [per the docs](https://xstate.js.org/docs/guides/typescript.html#config-objects), so we could add that later but it isn't in scope at this time.

stonecrop/src/registry.ts
```diff
@@ -9,22 +9,23 @@ export default class Registry {
...
+	constructor(router: Router, doctypeLoader: (doctype: string) => Doctype | Promise<Doctype> = undefined) {
 		if (Registry._root) {
 			return Registry._root
 		}
 		Registry._root = this
 		this.name = 'Registry'
 		this.router = router
 		this.registry = {}
-		this.schemaLoader = schemaLoader
+		this.doctypeLoader = doctypeLoader
```
agritheory, line 20: This is a better name, good change...

stonecrop/src/registry.ts
```diff
@@ -9,22 +9,23 @@ export default class Registry {
...
 	}
 
 	loadDoctypeSchema(doctype: Doctype) {
-		if (doctype.schemaLoader) {
-			return doctype.schemaLoader()
-		}
+		// if (doctype.schemaLoader) {
+		// 	return doctype.schemaLoader()
+		// }
```
agritheory: But you'll want it here too

Alchez: So this specific function is only responsible for loading a schema, given a doctype instance. I think this utility is no longer relevant since the schema is already available in the doctype instance itself, so we should be able to just remove this function entirely. What do you think?

agritheory: That makes sense, the registry is expecting a complete doctype so schema shouldn't be loaded separately. Let's remove it.

PR #40, opened 2022-10-26
[atable] add expandable list row story

### LINKED ISSUE #36: Expandable Row component
- [x] Inject a `div` inside the current row after the `<td>`s and and `<AForm>` inside that.
- [x] Final story should include Table > Form > Table > Form implementation (List view, Record, Child Table, detailed child view)
- [x] Key binding to toggle expand/ collapse (probably CTRL+G)

Alchez:
Closes #36.

### DIFF
--- a/aform/histoire.config.ts
+++ b/aform/histoire.config.ts
@@ -1,8 +1,0 @@
-import { defineConfig } from 'histoire'
-import { HstVue } from '@histoire/plugin-vue'
-
-export default defineConfig({
-	plugins: [HstVue()],
-	setupFile: 'src/histoire.setup.ts',
-	storyIgnored: ['**/node_modules/**', '**/dist/**'],
-})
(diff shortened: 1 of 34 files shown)


agritheory:
@Alchez Looks like a missing style import from AForm into the story
image

Alchez:
@agritheory, I wasn't able to do an import for the stylesheet for ATextInput, and I'm not sure why the scoped CSS wasn't included with the ATextInput component itself. But for now, I've duplicated the CSS inside the story so that it renders correctly.

agritheory:
@Alchez It's not just the styling for the input, we also need the styling for AForm - specifically that these should be narrower width than they are appearing and flexing inside the AForm fieldset.
Also, the header and first row are repeating (?) below:
image

agritheory (approved):

PR #45, opened 2022-12-20
[code-editor] create new Monaco editor project

### LINKED ISSUE #38: Code Editor Component
- [ ] Code Editor component, compatible with AForm API (https://github.com/bazingaedward/monaco-editor-vue3)

Alchez:


agritheory:
Can we make this 100% width by default?
image

What languages are supported? I can't the "language" input in Histoire to work.

agritheory:
@Alchez Can you also change the package name to `code_editor` - with plans to have an EditorJS component at some point and with currently undefined requirements, the package name of 'editor' is too ambiguous.

Alchez:
@agritheory 

> Can we make this 100% width by default?

Yeah, I've made some changes so that the editor always picks up the full area of its container.

> What languages are supported? I can't the "language" input in Histoire to work.

I think the Histoire controls are broken a bit, but I was able to make it work by:
- Creating a new preset 
- Setting the `language` and `initialValue`
- Saving the preset
- Reloading the story

> Can you also change the package name to code_editor - with plans to have an EditorJS component at some point and with currently undefined requirements, the package name of 'editor' is too ambiguous.

Done

<hr>

Should I add another story that uses this code editor component in AForm?

agritheory:
@Alchez `code_editor`, rather than `code-editor`. Sorry for being pedantic, but it will be harder to change later.

code_editor/src/components/ACodeEditor.vue
```diff
@@ -0,0 +1,41 @@
...
+	editor.defineTheme('agritheory', theme)
+	editor.setTheme('agritheory')
+
+	// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+	editor.create(container.value as HTMLElement, editorOptions)
+})
+</script>
+
+<style scoped>
+#container {
```
agritheory: I am really uncomfortable with this global id. Is this required by Monaco? Can we refactor it to a different selector?
Related, the height styling doesn't really make sense to me: when I change it to, for example, '30vh' in the story, there is still ~50% more code editor space that appears in on scroll. When I change it to '100vh' there is still more code editor area below the scroll.

agritheory (changes requested):

code_editor/src/components/ACodeEditor.vue
Alchez: @agritheory, turns out Monaco will add some view height beyond the last line. There's an option we can use to either disable it for the component, or just for the story.

agritheory: @Alchez Let's disable it. It's going to scroll appropriately

Alchez: Alright, I've disabled it for the entire component in the last commit. If we want the user to allow for that, I can move it to the story instead.

agritheory (approved):

PR #46, opened 2022-12-28
[examples] Docbuilder theming, page actions, and state editor

### LINKED ISSUE #47: [NodeEditor] StateEditor Ribbon and layout
Currently the buttons in the StateEditor component stack above the component. I'd like to see them in a ribbon that spans 100% width of the component and adds buttons. This should be ~1.15em tall, similar to the default row height in ATable

- Add a "Center" button that resizes the display and centers it for the user
- Double clicking on a node or transition should allow you to rename it
- When a transition is selected, it should offer options to insert a new node
- Make StateEditor compatible with v-model so have an easy API to automatically populate the Events table in the doctype builder example

### LINKED ISSUE #50: [Desktop] ActionSet Overflow issue
image
I think we want this to be fixed height and have an overflow of hidden

johnhkelley1:


agritheory:
@johnhkelley1 I have lots of ideas about this but want to focus on reasonably sized changes to this specific PR. Let's put the workflow editor inside a fieldset. I am going to make a separate commit on this branch with a better story that aligns the data in the example.

agritheory (approved):
Merging so we can start working on downstream problems in parallel.

PR #51, opened 2023-01-05
[aform] add slot for fieldset for manually supplier components

agritheory:
This looks to be useful when we want to wrap a non-AForm component with a fieldset, like StateEditor. 

@Alchez Does `SchemaTypes[]` need to change to be optional? Or conditionally optional?


agritheory:
@Alchez Can you review this when you get a chance?

Alchez (changes requested):
@agritheory, I think the `schema` prop can become optional in the code, but that behaviour probably needs user documentation to avoid confusion.

aform/src/components/form/AFieldset.vue
```diff
@@ -4,7 +4,9 @@
 			{{ label }}
 			<CollapseButton v-if="collapsible" :collapsed="collapsed" />
 		</legend>
-		<AForm v-show="!collapsed" :schema="schema" :data="formData" />
+		<slot>
+			<AForm v-show="!collapsed" :schema="schema" :data="formData" />
+		</slot>
```
Alchez: I think the `slot` should also expose the `collapsed` prop, since that's not available for the user

```suggestion
		<slot :collapsed="collapsed">
			<AForm v-show="!collapsed" :schema="schema" :data="formData" />
		</slot>
```

agritheory: I tried that and it doesn't work. I need to find the docs that say this, but it's not allowed. I couldn't come up with an alternative that I liked, maybe it needs a wrapper?

Alchez: This was the docs I was referring to: https://vuejs.org/guide/components/slots.html#scoped-slots. Not sure why it wouldn't work though.

agritheory: This looks like its worth a try.

PR #54, opened 2023-01-16
[atable] allow tables to occupy full-width

### LINKED ISSUE #48: [ATable] ATable API(s) for full width
We want to be able to tell a table to size its columns to fill the available space. This should be a top-level API to progressively hide columns as horizontal space is reduced (or not) as well as respecting the "width" parameter to mean "min-width".

Alchez:
Closes #48


Alchez:
@agritheory, this is almost ready to go, but for some reason, the classes for the index columns (both header and data) are not being loaded. Can you have a look on your side and see what might be happening?

atable/src/components/ATable.vue
```diff
@@ -14,16 +14,17 @@
...
 						:col="col"
 						spellcheck="false"
 						:rowIndex="rowIndex"
 						:colIndex="colIndex + (tableData.zeroColumn ? 0 : -1)"
 						:component="col.cellComponent"
 						:style="{
-							textAlign: col?.align?.toLowerCase() || 'center',
+							textAlign: col?.align || 'center',
 							minWidth: col?.width || '40ch',
+							width: tableData.config.fullWidth ? 'auto' : null,
```
agritheory, line 27: @Alchez `min-width`, then `width: auto` works perfectly. Your code for freezing the width of the first column if perfect. Try resizing the window in the full width story.

atable/src/components/ATableHeader.vue
```diff
@@ -1,7 +1,7 @@
 <template>
 	<thead v-if="columns.length">
 		<tr class="atable-header-row" tabindex="-1">
-			<th v-if="tableData.zeroColumn" id="header-index" />
+			<th v-if="tableData.zeroColumn" class="header-index" />
```
agritheory: @Alchez  I would love to find another selector for this or even do the same v:bind sort of trick you did elsewhere. I was looking for documentation to use `ref` as a style selector but came up empty. I think a class that isn't designed to be exposed is going to end up being leaky CSS.

agritheory (changes requested):
One thing left to do RE `header-index`, not sure how to tackle it

atable/src/components/ATable.vue
Alchez: Oh, I think I misunderstood how `min-width` and `width` work. That's a good fix!

atable/src/components/ATableHeader.vue
Alchez: @agritheory I was able to move this back to an ID selector with the changes you made, and the zero-column width worked as expected. Since the CSS in that file is scoped, should we still look at alternatives?

agritheory: Yeah, that should be fine.

agritheory (approved):

PR #72, opened 2023-02-09
[aform][atable] v-model interface for AForm and ATable

### LINKED ISSUE #19: [AForm] Make AForm compatible with v-model

johnhkelley1:


aform/src/components/AForm.vue
```diff
@@ -1,26 +1,30 @@
...
 
 <script setup lang="ts">
-import { ref } from 'vue'
+import { ref, computed } from 'vue'
 
 import { SchemaTypes } from 'types'
 
 const props = defineProps<{
+	modelValue: SchemaTypes[]
 	schema: SchemaTypes[]
```
Alchez, line 22: Should we ask the client to [specify a name for `v-model`](https://v3-migration.vuejs.org/breaking-changes/v-model.html#v-model-arguments)? That would help avoid duplicate props.

It will have to be documented, but instead of `v-model` and `update:modelValue`, the client would have to use `v-model:schema` and `update:schema`, which would be more explicit.

agritheory: Changes to schema from a component don't need to emit. We want it to be reactive so that changes from outside the component are propagated, but there's no change to a component's schema that should bubble back up.

agritheory:
Closes #53 and #19

agritheory (approved):

PR #74, opened 2023-02-10
[atable] allow top-level readonly for ATable

### LINKED ISSUE #60: [ATable] Read-only API
Cascade a top level API on the Table to its columns and components to make them read-only

Alchez:
Closes #60.

### DIFF
--- a/aform/src/components/form/ADate.vue
+++ b/aform/src/components/form/ADate.vue
@@ -1,5 +1,6 @@
 <template>
 	<div
+		v-if="!readonly"
 		:event="event"
 		:colIndex="colIndex"
 		:rowIndex="rowIndex"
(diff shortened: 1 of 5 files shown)


Alchez:
@agritheory Since the tables are already read-only, I've applied it to modals instead. Not sure if we want to suppress modals, or show modals with the underlying component as read-only?

agritheory:
@Alchez The modal component needs to receive a read only prop and then do the right thing with it, which would be component specific. A datepicker doesn't need to show the calendar if it's read only - no modal, but some other component - maybe the quantity component that show qty and uom at the top level but carries a qty, uom, conversion ratio, stock uom and stock qty inside the component - that might use the modal but since it has a read only prop it would make all the fields inside it read only.

Alchez:
@agritheory, that makes sense. I'm currently having the stories pass the modal props via the schema, since that's how we're defining and loading up the modals anyway. Should I also add it as a prop to the table components as well? Or should we wait for a use-case?

agritheory (approved):

PR #98, opened 2023-09-06
[aform] adate refactor

Alchez:


agritheory:
@Alchez I'm stuck on a couple of issues here:
1. The emits aren't correct - they should be using the same nomenclature as other form components.
2. The `selectDate` function does not seem to be working correctly, or at least it's not emitting a value. Might be related to 1.
3. Styling - we need to assume that this picker doesn't _only_ occur in a dropdown/table context.

Can you take a look at this and give me a debrief of the mistakes?

Alchez:
@agritheory

The emits should now match up with the other form components. Checkbox still seems to be using a slightly different emit event but I haven't changed it in this iteration yet. The `selectDate` function signature was incorrectly mapped on the template, but it should be emitting a value now (tests are passing though).

As for styling, I'm not too sure how to approach that one. Let's discuss on a call tonight?

agritheory:
@Alchez Keyboard navigation
- [ENTER] should select the currently highlighted date
- [TAB] should blur and propagate so as to select the next field (outside of ADatePicker)
- [SHIFT] modifier should have the same behavior as [TAB] and [ENTER]
- [ESC] should have the same behavior as [TAB]
- [DEL] Should unset the selected date value

agritheory:
@Alchez Let's refresh this one

Alchez:
@agritheory, there seems to be a dependency resolution issue with `@stonecrop/utilities` in both `aform` and `atable`, but only for the GHA environments. The tests are working and passing as expected locally for me.

I am tracking an [open issue](https://github.com/vitest-dev/vitest/issues/5486) that was active a couple of weeks ago, so I'll try the workarounds mentioned in that.

agritheory:
@Alchez Some issues I noticed in the ADatePicker story:
- page up/down increments the month for both interfaces in the story
- enter should select the current date

In the ADate story I expected to see the ADatePicker interface show below the input. It is also very wide, is that because it's not in a fieldset?
I would expect typing (or pasting) in the ADate field to update the values in ADatePicker, kind of like a mask

@crabinak Design dilemma: how do we visually communicate "today", "selected" and "cursor" in this interface without relying on a theme color? 

image

crabinak:
@agritheory I'll take a look but, I'm thinking something along the lines of:

Today: Filled background.
Selected: Dashed line border.
Cursor: Solid line border. 

Or vice versa with selected and cursor, I just think of the marching ants border of selection tools which is why I went for dashes. Dashes might look terrible, so it might be a thicker or thinner border instead. I'll try a few options and see what works.

Alchez:
@agritheory 

> * page up/down increments the month for both interfaces in the story

I can replicate this on my end, but I'm not sure why it's happening. I'm guessing since the pgUp event listener is loaded up on every cell, it gets triggered for both tables everytime it's pressed? 

> * enter should select the current date

I've added a commit that sets up this behaviour.

> In the ADate story I expected to see the ADatePicker interface show below the input.

I looked into the `<input type="date">` documentation, and as far as I can tell, the extent to which you can override the datepicker is just styling. I couldn't see any way to replace the datepicker with our version.

> It is also very wide, is that because it's not in a fieldset?

I think that might be true. I'm using a simple `<input>` tag without any special styling. There is an import for aform stylesheets, but those might only apply for inputs within a fieldset?

> I would expect typing (or pasting) in the ADate field to update the values in ADatePicker, kind of like a mask

Technically this does work for the ADate input, but if we want ADate to work with ADatePicker, then I think we'll have to design something. Maybe an `<input>` tag with the `type="date"` attribute? And then we insert the modal ourselves?

agritheory:
@Alchez The UX I think we want is based on the user tabbing versus clicking into a date field. If they've clicked on it the field we can assume they want a datepicker interface and should only evaporate it after they blur from the field. On the other hand, when they tab into the field, we can assume hands stay on the keyboard and we can respect their text-based input.

agritheory:
> I can replicate this on my end, but I'm not sure why it's happening. I'm guessing since the pgUp event listener is loaded up on every cell, it gets triggered for both tables everytime it's pressed?

I don't think we want page up/down listeners on every cell, is this based on the utility composable?

crabinak:
@agritheory I just updated the style on the date picker. I bolded and underlined the current day so that is it emphasized but does not overpower/compete with the selected day. Let me know if something isn't looking right.

Additionally, all days that have already passed could be given a lighter gray color, but I wasn't sure if the user would still be able to select these days (in which case graying them out might give them the impression they are unselectable).

aform/src/components/form/ADate.vue
```diff
@@ -1,277 +1,94 @@
...
-			</tr>
-		</table>
+	<div>
+		<input
+			ref="dateRef"
+			type="date"
+			:id="uuid"
+			:disabled="readonly"
+			:required="required"
+			:value="inputDate"
```
Alchez, line 9: @agritheory I had to move the map from `v-model` to `value` since the two-way sync was causing some issues with the `v-model` setup in the story. Let's discuss this change.

Alchez:
@agritheory I was looking to use [`useFocusTrap`](https://vueuse.org/integrations/useFocusTrap/) to avoid the pgUp and pgDown events to trigger on both the datepickers in the AForm story, but that didn't work as intended.

I tried implementing it in just the Histoire story as well as trying to apply it to the `ADatePicker` component (although I'm not sure if this would have worked as intended since the focus trap would have been applied twice?).

Should we switch approaches and apply/remove keyboard navigation listeners conditionally? Maybe using [`useFocusWithin`](https://vueuse.org/core/useFocusWithin/)?

What do you suggest?

agritheory:
@Alchez I think its worth experimenting with `useFocusWithin`, it sounds promising.

PR #105, opened 2024-04-16
[stonecrop] publish packages to NPM

### LINKED ISSUE #100: Publish to NPM
GitHub packages might be worth it for top secret stuff, but I don't want to have to install this into a production environment or Frappe Cloud.

Alchez:
Closes #100.

---

Additional notes:
- I tested it locally, and I was able to publish the packages to my [personal account](https://www.npmjs.com/~alchez).
- Since the package namespace has changed and they aren't published on NPM yet, they won't be available for local install. Until they're published, a local test would require managing dependencies across the project using `yarn link` (to register a package locally) and `yarn link <project_name>` (to use the registered package locally).
- If you're testing this locally, you would need to setup an access token in your NPM account (https://www.npmjs.com/settings/USER/tokens) and export it in your local env as a variable called `NODE_AUTH_TOKEN`.
- For production releases, we would need to add an `NPM_TOKEN` secret in this repo's [action secrets](https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions#creating-secrets-for-a-repository). Since we used to use Github's package registry, this was automatically available as an env variable.


agritheory (approved):
Looks good. I added the secret to the repo.

PR #113, opened 2024-04-18
[aform] add login utility component

### LINKED ISSUE #110: [aform] Login component and example
Unlike other components in this repo, this one is not schema-driven. It will exist here mostly as a utility and to have consistent theme-ability. I propose using the shadcn Vue for the general template but harmonize our default theme and the same border styles as AForm. 

https://www.shadcn-vue.com/examples/authentication.html

Alchez:
Closes #110.

---

**ToDo:**
- [x] Confirm styling generally matches the default theme
- [x] Create a login example story


Alchez:
@agritheory @crabinak I've added a login component that's similar to the shadcn-vue template (but it's made for login rather than creating an account).

I've setup some basic styling based on the Tailwind classes setup in their template, but I think someone else should handle the stylesheets here :sweat_smile:

crabinak:
@Alchez I can handle the stylesheet on this.

Alchez:
> @Alchez I can handle the stylesheet on this. 

Thank you!

Alchez:
@crabinak @agritheory I've created a merge commit that removes the dist folders from the diff, so the changes should be easier to read now.

crabinak:
@Alchez @agritheory I have added styling to this login. I'm importing styles from the default.css file to avoid duplication and maintain consistency, but let me know if this violates the monorepo design pattern at all. 

I've also setup an animated loading icon to appear when the user clicks the login button and bound it to isLoading. This uses Google Material Icons, but let me know if a different icon library is preferred.

agritheory (approved):
This looks great. @crabinak Loading animation is very nice. I think we might explore some more custom work there soon.

PR #114, opened 2024-04-18
[stonecrop] use new v-model API

### LINKED ISSUE #112: [project] New preferred API for v-model
https://vuejs.org/guide/components/v-model.html

Which make we want to refactor the appropriate components to use a composable-style API for this kind of thing. The argument feature is pretty cool.

Alchez:
Closes #112.

--- 

**Todos:**
- [x] Figure out edge cases of v-model that have more complex setups or processing.
- [x] Figure out migration cases to move stonecrop components to a composable-style API


Alchez:
@agritheory I've only updated the cases where the switch from the old way is easiest. There are still a few cases where the update is a bit more complex which we can look at together (for example, the `v-model` usage in `aform` uses a nested setup which as far as I can tell isn't a straightforward move).

Regarding the refactor to a composable style, we can also talk through some examples of what you're expecting.

agritheory (approved):
As we talked about, these dist diffs are really discouraging

Issue #118, opened 2024-04-22
[themes][aform] ATable headers seems heavy compared to AForm field labels 


agritheory:
As far as communicating importance to the user, the field level label and column label are equivalent, but they sure do not feel that way. I suspect the the field label needs to be "heavier" (maybe in font weight, maybe in "impact") and the table column header styles are by comparison too dark and heavy. 

![image](https://github.com/agritheory/stonecrop/assets/18033666/88b7d39f-54f9-42cd-a4b8-350dcb63bac4)

![image](https://github.com/agritheory/stonecrop/assets/18033666/d2143b70-519a-406f-9bd1-e5c80d2e3971)

Another thought about stylistic consistency: I feel like the distance between the top and bottom of the field label border should be the same as the default line height in the table.

agritheory:
@crabinak Interested to hear your thoughts on this one and any adjacent issues you notice.

crabinak:
@agritheory I'll take a look. I agree on field padding. The field labels can probably bump up out of the box and get a bolder treatment. I'll experiment and see what works.

PR #119, opened 2024-04-24
updated form labels to match table heading. consolidated similar styl…

crabinak:
I updated the form field headers to more match the table headings. I also consolidated a lot of duplicate styles from these form elements into a single file they can share. I also replaced scoped styling with class-level styling on these components. I also bumped up the input label text color styling from gray--80 to gray--60 in default.css as I believe that is what the table headers are using.

Let me know if this moves closer to resolving issue #118. 

image


crabinak:
@agritheory I have updated the table and form, heading styles to match within the expandable table. Both headings are now a bit lighter. I moved most of the table styles up into the root to make them easier to access. 

One thing I noticed in the form is that there are paragraph elements used to show error message feedback beneath each field. This provides for some awkward gaps between form elements vertically (this is how it currently is). My original thought was to toggle their display mode when they are shown, but this would cause uneven spacing for elements within the same row.

My thought on these is to have all error feedback display below the form element and the individual paragraph elements can be removed/pushed down to under the form. I have a mockup of how it would look attached. This would close up the spacing on the form elements quite a bit. 

error-messages-collated

crabinak:
@agritheory I updated the form to match the styles we had discussed. I also moved the error messages to appear over the fields (populate an error to visualize this). These will sit on the field like the label currently does, but right aligned. This eliminates the spacing issue below the fields. 

EDIT: @agritheory I'm adding the photos here. I also forgot to mention I split this into a three-column layout instead of two for fields as there is more than enough room for each field. 

Aform:
aform

Atable (with error message in red)
aform-table-expanded

agritheory:
@crabinak We talked about this a few weeks ago, but I think we want to remove the background color from the table header (red arrow) and adding it to the left border of the table (like green box). I'm also wondering if it makes sense to center the top left corner of the table like blue box against the header text (blue box). 

expandable_table_changes

crabinak:
atable-header

@agritheory I updated the header and border colors here. I was unsure what this meant:

> I'm also wondering if it makes sense to center the top left corner of the table like blue box against the header text (blue box).

agritheory:
> @agritheory I updated the header and border colors here. I was unsure what this meant:
> 
> > I'm also wondering if it makes sense to center the top left corner of the table like blue box against the header text (blue box).

I'll try to explain, but I'm not great at this and that's why I need your help.

I think the thicker border on the left of the fieldset (green box in the marked-up image) is really nice. I think it might be a good left border for the table component too. Let's start there, but maybe we'll iterate on it some more.

crabinak:
@agritheory I went ahead and applied the border to the table as well. This one is slightly darker than the form border, but it kind of helps reinforce hierarchy. 

atable-table-border

agritheory:
@crabinak This looks great and I see your point about hierarchy. I think we should live with this for a while and remain open minded about changing it if it becomes distracting when in use.

PR #124, opened 2024-05-02
[project] upgrade VueUse to v10

### LINKED ISSUE #120: [utilities] VueUse upgrade limitation
VueUse released [v10](https://github.com/vueuse/vueuse/releases/tag/v10.0.0) last year, but we have a breaking change with [`useElementVisibility`](https://vueuse.org/core/useElementVisibility/#useelementvisibility) while working with keyboard navigation. In versions <10, they were using a scroll event handler, but they updated the internals of the composable to use [`IntersectionObserver`](https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver) instead. This is causing our keyboard navigation that relies on the composable to stop working.

Until we figure out a way to use the new APIs, we're blocked from upgrading to v10 if ever it's required.

Alchez:
Closes #120.

### DIFF
--- a/common/changes/@stonecrop/stonecrop/feat-vueuse-upgrade_2024-05-02-09-09.json
+++ b/common/changes/@stonecrop/stonecrop/feat-vueuse-upgrade_2024-05-02-09-09.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@stonecrop/stonecrop",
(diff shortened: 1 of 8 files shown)


Alchez:
@agritheory This one requires an upstream bug report like you suggested, but I'm having issues trying to get an environment working. Let me try work on that.

Issue #49, opened 2023-01-05
[ATable] Draggable/ Resizable columns in ATable


agritheory:
A header component that can slot into ATable

https://vueuse.org/core/useDraggable/

This should be exposed as configuration / schema API at the ATable level

bhattdevarsh:
Trial conclusion: `useDraggable` is not the solution for this scenario.

bhattdevarsh:
`useDraggable` is possible but will require muti-level logical changes to make sure other element like cell and header are aligned correctly and adapt the drag event. Additionally there is something called `axis` that can be programmed to move element only in preferred direction, but that is available in `v10.9.0` and not on `v9.13.0`.

![image](https://github.com/agritheory/stonecrop/assets/58166671/8efde448-dbc6-42b5-be34-eca09551c37f)

https://github.com/agritheory/stonecrop/assets/58166671/fc3b2968-a161-488e-8424-28a16a84c870

Alchez:
@agritheory @bhattdevarsh This is also related to #120. The PR that I've raised should allow us to start using v10 APIs.

Alchez:
@bhattdevarsh this one can move ahead with v10.9.0

bhattdevarsh:
Closing this as the PR is merged.

PR #126, opened 2024-05-07
[aTable] allow user to move table rows

bhattdevarsh:
https://github.com/agritheory/stonecrop/issues/129


bhattdevarsh:
The Table rows are draggable as required, but the feature is not complete yet:

After the drag event, the table row is being placed correctly and sorting it as intended, but the `tableData` is getting refreshed due to which it appear as it is not doing anything, attaching video which demonstrate row 1 moved to row 10 and rows between 2-9 are shifting 1 row above and then the table data is refreshing itself in console log.

https://github.com/agritheory/stonecrop/assets/58166671/d4ad223a-2a07-407f-b408-c7ad0b5d4fa6

agritheory:
@bhattdevarsh This is not the feature associated with #49 which was intended to be able to resize columns. That said, this is a nice-to-have, so I suspect we'll end up completing it but associated with another ticket.

atable/src/components/ARow.vue
```diff
@@ -43,6 +44,20 @@ const tableData = inject<TableDataStore>(props.tableid)
 const rowEl = ref<HTMLTableRowElement>(null)
 const numberedRowWidth = tableData.numberedRowWidth.value
 
+const { x, y, style } = useDraggable(rowEl, {
+	initialValue: { x: 5, y: (props.rowIndex + 1) * 21 },
```
Alchez, line 48: Don't use [magic numbers](https://en.wikipedia.org/wiki/Magic_number_(programming)). Is the `21` supposed to be the row height? Can we compute it instead?

atable/src/components/ARow.vue
```diff
@@ -43,6 +44,20 @@ const tableData = inject<TableDataStore>(props.tableid)
 const rowEl = ref<HTMLTableRowElement>(null)
 const numberedRowWidth = tableData.numberedRowWidth.value
 
+const { x, y, style } = useDraggable(rowEl, {
```
Alchez: Without passing a `containerElement`, I think the composable is letting you drag the row outside the bounds of the table, which is causing some issues with the row data (random `undefined` rows)

Alchez (changes requested):
@bhattdevarsh I tried the following methods locally to update the table but with no luck:
- Using a `key` attribute on the `ARow` and `ATable` top-level elements (via an emit) to [force a re-render](https://blog.logrocket.com/correctly-force-vue-component-re-render/#key-changing-technique) whenever a re-arrange occurred.
- Manually calling `forceUpdate` on the Vue component, but that didn't help either.

Alchez:
Closing in favour of #277, since I can't push changes here

PR #134, opened 2024-06-08
Beam css leak fix

crabinak:
I fixed the style leak that was occurring. The HTML element was being styled in beam.css. I have removed it and updated all rem values based on 1em = 16px instead of 1em = 10px  they were using before. Let me know if you see any other issues occurring here.

### DIFF
--- a/beam/themes/beam.css
+++ b/beam/themes/beam.css
@@ -9,11 +9,6 @@
 	--brand-color: #202a44; /* brand color, default navy */
 }
 
-html {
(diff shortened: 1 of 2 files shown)


agritheory (approved):

Alchez:
@crabinak can you please run `rush change` locally for this branch? It'll prompt you for a changelog message, which should help resolve the tests here

Alchez (approved):

PR #137, opened 2024-07-08
[project] avoid recursive rendering issues

Alchez:
Changes:
- Remove `v4` uuid calls in components to avoid forcing re-render, leading to recursion errors
  - `Uncaught (in promise) Maximum recursive updates exceeded in component <ATable>. This means you have a reactive effect that is mutating its own dependencies and thus recursively triggering itself. Possible sources include component template, render function, updated hook or watcher source function.`
- Replace function calls with computed properties wherever applicable
- Replace `key` values from indexes to primary key values ([reasoning](https://stackoverflow.com/a/44531545)).


agritheory (approved):
Nice changes, as discussed on the call, we can improve some aspects of this but it isn't in the critical path right now.

PR #141, opened 2024-07-25
[atable] pass additional context while formatting table cell value

### LINKED ISSUE #139: [ATable] inject collect and inject variables into ATable
```js
                {
                    "component": "WorkQueueDescription",
                    "name": "description",
                    "fieldtype": "Data",
                    "label": "Description",
                    "required": false,
                    "width": "50ch",
                    "format": "(cell) => {console.log(cell)}"
                    "format": "(row, cell) => {console.log(row, cell)}"
                    "format": "(table, row, cell) => {console.log(table, row, cell)}"
                },
```

This will allow formatting or masking of any of these variables. I think this API is OK to be inferred based on the number of arguments provided.

Alchez:
Closes #139.

---

image

### DIFF
--- a/examples/atable/default.story.vue
+++ b/examples/atable/default.story.vue
@@ -16,10 +16,10 @@
 </template>
 
 <script lang="ts" setup>
+import type { CellFormatContext, TableColumn } from '@stonecrop/atable'
 import { ref } from 'vue'
 
 import data from './sample_data/http_logs.json'
-import type { TableColumn } from '@/types'
 
 const columns: TableColumn[] = [
 	{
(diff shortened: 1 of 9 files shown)


examples/atable/default.story.vue
```diff
@@ -62,10 +62,8 @@ const readonly_columns: TableColumn[] = [
 		type: 'Data',
 		align: 'left',
 		edit: false,
-		width: '35ch',
-		format: (value: { title?: string; value?: any }) => {
-			return value.title
-		},
+		width: '40ch',
+		format: '(value, { row }) => `${value.title} (IP: ${row.ip_address})`',
```
Alchez, line 66: @agritheory this is the example of the string function format I've used, but the full signature is:

```
format: "(value, { table, row, column }) => return { console.log(table, row, column) }"
```

You can also do:

```
format: "(value, context) => return { console.log(context.table, context.row, context.column) }"
```

PR #143, opened 2024-08-14
[atable] allow pinned columns in ATable

### LINKED ISSUE #142: [ATable] Freeze columns

Alchez:
Closes #142.

I've created a new "pinned columns" variant in the example ATable stories that applies a basic version of `position: sticky` to just the index column (`class="sticky-index"`) for now, with a base for applying to arbitrary table columns (`class="sticky-column"`) as well.

But this needs to be expanded to allow columns to be dynamically pinned based on the provided schema config.

I don't think this can be a CSS-only solution anymore, but I'll defer to @crabinak.


crabinak:
@Alchez This looks like a good start in the right direction. I came up with a quick demo implementation using sticky positioning, flexbox ordering, and dynamically positioning columns based on column width when sticky. It is just a matter of giving a cell the "sticky" class. 

[Sticky columns demo](https://testing.crabinak.com/html_demos/sticky-columns/)

I created a new feature branch off this and am working on implementing this into atable to see if it will work.

agritheory:
@crabinak This demo looks great, I think _just_ clicking to make something sticky probably isn't the right interaction, especially since it will likely interfere with other editing features. I think we want this to be defined at the schema level, which means we can expose the feature into user-space when required.

agritheory:
@crabinak Let's make sure the specimen also allows for resize. We don't need to support making columns sticky without re-rendering.

agritheory:
@crabinak @Alchez Can I get you guys to give this a "ready for review" blessing?

Alchez:
@crabinak @agritheory Seems like the pinned columns feature is also available for treeviews. Should we let this go as is, or add a failsafe in case someone accidentally applies pinning to tree-view lists?

agritheory:
@Alchez What is the UX argument against allowing it? You can imagine a financial statement with grouped rows and a pinned "account" column.

Alchez:
@agritheory @crabinak I think my only concern is UI for tree-views, since the pinned columns would need to work with both index and row-expansion columns.

agritheory:
@Alchez Let's put them in a story and see how it goes.

Alchez:
@agritheory @crabinak I've added a variant in the tree stories for pinned columns

crabinak:
@agritheory @Alchez I added the pinned column feature to the tree view. Since rows are added and removed, I had to add a Mutation Observer to the table to detect and readjust newly added rows/cells.

PR #144, opened 2024-08-16
[beam] build styles

Alchez:


Alchez:
@agritheory this should fix the list styling for https://github.com/agritheory/beam/pull/121 and https://github.com/agritheory/beam/pull/154. Turns out they weren't being included in the package at all.

agritheory (approved):

Issue #146, opened 2024-09-06
[beam] navbar improvments


agritheory:
- Remove icon
- Float / sticky navbar to top
- reduce height 
- Not sure why it's not full width here

Let's handle improvements to the "Done" button in another pass, once we have some more functionality outlined 

![image](https://github.com/user-attachments/assets/21620399-e660-4c36-a348-96138f2a757e)

crabinak:
@agritheory This shouldn't be a problem. Unfortunately, I am now effectively blocked from working on Beam until I resolve this issue I keep having whenever I try to run rushx dev commands from the examples directory. 

I've spent some time debugging the issue and my best guess is when I run `rushx dev:aform` from the examples directory, it cannot find the vite.config.ts file it needs to reference (maybe it is still searching the examples folder when it should be searching the aform/beam/directory-it-switched-into). It is failing to load in the vue or histoire plugins as it seems to be falling back to a default vanilla config. I've used the `--config` flag to manually direct it to the correct config file, but it still ends up with the same error. _(this is all just my best guess from using console.log in the histoire code)_

I've run `rush purge && rush update && rush rebuild` several times as well as manually removing all node_modules I could locate. I have also tried a fresh clone of the repo and this still occurs. 

What I do not understand is why I am the only one running into this issue even when I clone fresh from the repo. Could there be something external to the repo I am missing? 

I've run out of ideas and was seeing if anyone might have an idea what is going on here? We can discuss this on our call Monday and I'll set it up so I show this issue, but I wanted to mention that I am at a roadblock at the moment. Here is another screenshot of the error I keep running into.

![plugin-error](https://github.com/user-attachments/assets/4922fb0b-34e5-4a5f-a341-84984fff195e)

Alchez:
@crabinak Just to confirm, is this only happening on running `rushx dev:aform` or also on `pnpm run dev:aform` (which had worked for you before as a stop-gap)?

I tried to replicate this locally by changing the Vite-Histoire config, and I think you're correct that this must be an issue with Histoire's Vue plugin registration. But I'm not sure why that would fail like this.

> What I do not understand is why I am the only one running into this issue even when I clone fresh from the repo. Could there be something external to the repo I am missing?

The only difference I can see is that you're running stonecrop on native Windows, whereas I'm using native Ubuntu. I think Tyler might be running WSL2 too, instead of native Windows. Could it be a paths issue in Rush or Histoire?

crabinak:
[...]

PR #147, opened 2024-09-10
reduce navbar height, remove icon, made sticky navbar

crabinak:
@agritheory I made the updates outlined in #146 . It looks like the navbar is full width to me, but let me know if you are still seeing that issue. 

_I had to add a workaround on one of the Histoire wrapper classes as `overflow:auto` breaks the `position:sticky` functionality of children elements. I put this in the beam.css file and it shouldn't cause any conflicts since the Histoire class name is targets is hyper specific._

beam-navbar-fix

### DIFF
--- a/beam/src/components/Navbar.vue
+++ b/beam/src/components/Navbar.vue
@@ -1,8 +1,5 @@
 <template>
 	<nav class="beam__navbar">
-		<slot name="icon">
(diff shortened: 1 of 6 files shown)


agritheory (approved):

agritheory:
@crabinak Looks great, thank you !

PR #149, opened 2024-09-13
[beam] add onscan.js dependency

Alchez:
Required for https://github.com/agritheory/beam/pull/167.

---

This PR adds [`onscan.js`](https://github.com/axenox/onscan.js) as a dependency for Beam to allow hardware barcode scanners to be used.

The Beam package now has the following behaviour:
- The `ScanInput` component now attaches an instance of `onScan` to the `window` object, and automatically adds event listeners for hardware scanning.
- The `ScanInput` component in Beam now accepts a `scanHandler` prop that gets passed to the `onScan` instance.
  - The existing emit on the scanned barcode is removed, since that's now expected to be implemented in the handler.

@agritheory I've removed `scanHandler` from the global `window` object. To run test simulations, we can use the composable instead: `window.scanHandler.scanner.simulate(document, '<barcodeString>')` -> `<instance>.scanHandler.scan('<barcodeString>', '<qty>')`.


agritheory (approved):

agritheory:
@Alchez I know we're trying to move fast here, but I think we want a small unit test for this: use the `simulate` API in onscan and ensure that it emits to the parent of the `ScanInput` component.

Alchez:
@agritheory I've added a test that tests for the implementation code to have been run on simulating a barcode scanner.

agritheory (approved):

agritheory:
@Alchez How should I simulate this in the browser console for demonstration purposes?

Alchez:
@agritheory So I've made the `ScanInput.vue` component in Stonecrop emit the `onscan.js` instance whenever it is mounted. What you can do is pick up the instance in `Beam.vue` and run a `simulate` on that.

```vue
<template>
	<ScanInput :scanInstance="registerInstance" />
</template>

<script>
const registerInstance = (instance) => instance.simulate(window, '<barcode>')
</script>
```

PR #151, opened 2024-09-16
fix: isolate story-specific Beam styles

### LINKED ISSUE #150: [package / beam] Latest release 0.28 has broken styles

Alchez:
Closes #150.

---

So there's a couple things that I think caused the issue in the Beam app:
- The Histoire class overrides added in #147 were being built as part of the main `@stonecrop/beam` package's styles, which was then bundled together with the other dependencies in `agritheory/beam`.
- For some reason (they say it's for security, but there's no documentation around it), Frappe added a [change](https://github.com/frappe/frappe/pull/6351) that blocks Jinja templates/CSS stylesheets from rendering if they have the `.__` string in it, which the class overrides do.

This PR tries to move the class overrides from the main Beam theme to just the histoire instance in `examples/beam`, which prevents them from appearing in downstream installations. I think I was able to retain the sticky functionality, but can you give it another look to confirm @crabinak?

PS: @agritheory I'm also sneaking in a small change (https://github.com/agritheory/stonecrop/commit/5dfe65804483e85a461b8aab48651b7c1b17e1c1) with plugin installation for the Histoire setups. Instead of manually registering the components, I'm moving them to use the install function for each package, which handles the component registration for us. Let me know if they need to move back to granular imports.

### DIFF
--- a/beam/themes/beam.css
+++ b/beam/themes/beam.css
@@ -190,7 +190,3 @@ body {
 		color: var(--primary-text-color);
 	}
 }
-/* workaround for Histoire wrapper breaking sticky functionality with overflow:auto */
-.__histoire-render-story:not(.__histoire-render-custom-controls) {
-	overflow: clip !important;
-}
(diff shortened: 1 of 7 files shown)


agritheory (approved):

crabinak:
@Alchez It looks like the sticky header is still working. Thanks!

PR #157, opened 2024-09-28
[beam] Added Beam Metadata Component

### LINKED ISSUE #145: [beam] Metadata/ Document component
image

Let's assume these items are required in a single document. We need to know what kind of document it is (header) and three or four configurable facts about it. Examples given in context on the BEAM repo.

crabinak:
@agritheory I added the Beam Metadata component here. I added a slot for the ListView. I wasn't sure what type of data would exactly be expected, so I made a generic "workOrder" object that passes in as a prop and populates a few of the work order fields as well as binds some styling to the "complete" value.


agritheory:
image

agritheory (approved):

Alchez:
@crabinak I think this one just needs a changelog

PR #158, opened 2024-09-29
Feat pinned columns ordering

crabinak:
@agritheory @Alchez Referencing #142 I managed to get this working I think. 

Since the cells needed width data, this could only be gathered after the component mounts, so I added some code in the onMounted hook on the ATable component that gets a ref to the table and traverses through the cells, setting their correct left position if they are pinned. There is also a section that then corrects the heading widths based on their columns widths. You can try pinning various columns. I also added a border to the right edge of the last sticky item in each row to help visualize the break between pinned and non-pinned columns. 

I am not sure if the user will be able to resize columns on the fly, but I put all this into a single function that could be called on a column resize event to adjust all cells to proper values if needed.

There is one small bug that occurs where if the scrolling parent container is smaller than the column width + left position, it will begin to scroll again (this is just how sticky positioning works in relation to parent containers). I'll look into a workaround for this as it might come up if every column happens to be sticky (I do not know if that is a possible situation that may occur). I think there might be an answer somewhere in using absolute positioning over sticky positions, but I'll have to explore that more. 

I'm requesting this merge into the feat-pinned-columns branch since I branched off that.

### DIFF
--- a/atable/src/components/ACell.vue
+++ b/atable/src/components/ACell.vue
@@ -15,7 +15,7 @@
 		@click="handleInput"
 		@mousedown="handleInput"
 		class="atable__cell"
-		:class="hasPinnedColumns ? 'sticky-column' : ''">
+		:class="pinned ? 'sticky-column' : ''">
 		<component
 			v-if="currentColumn.cellComponent"
 			:is="currentColumn.cellComponent"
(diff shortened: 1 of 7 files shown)


agritheory:
@crabinak Merging, works great. Let's solve the resize and pin combination in a separate iteration

PR #165, opened 2024-10-13
[beam] metadata refactor

agritheory:
This is a (overly) dramatic refactor of the metadata component to remove the addresses, add a slot and remove the 'items' slot/opinion.  There's a lot of good things about the original implementation that I would like to retain, but I'm not sure about the best way to do that; perhaps they should become independent components which would allow them to be reused outside of the metadata component as well.

### DIFF
--- a/beam/src/components/BeamMetadata.vue
+++ b/beam/src/components/BeamMetadata.vue
@@ -1,55 +1,33 @@
 <template>
 	<div class="beam_metadata">
 		<div class="beam_metadata_content">
-			<div class="beam_metadata_header beam_metadata--two-column">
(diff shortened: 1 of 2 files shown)


crabinak:
@Alchez @agritheory This looks good. 

One comment, (now that I have a better idea of the device size this will be on) I think we can probably drop that gray arrow between "shipping" and "receiving" as it is hidden on mobile view anyways and doesn't add much value to the interface.

agritheory:
@crabinak Let's talk about how to preserve some of the other styling helpers / semantic elements you have created. I think we're going to want them in the future.

crabinak (approved):

PR #171, opened 2024-10-18
Feat beam components creation

crabinak:
Here are a few components I put together based on some of the styles from the Metadata Component. I also refactored some styling/class names for consistency, using a BEAM_ELEMENT-NAME--MODIFIER pattern (I know some issues were raised that some classes had generic names like .btn which is now .beam_btn). 

Some of the components:

1. **BeamHeading:** Basically an H1 wrapper for beam headings. This just saves some time from having to type an h1 and include the heading classes each time. 
2. **SplitColumn:** A two column layout. Props can be used to designate the content alignment and justification on flex axis. The left column is left aligned by default while the right column is right aligned. I was not sure if each column needed individual control for content alignment. This will collapse into a column upon reaching mobile sizes and left align everything.
3.  **Arrow:** The Arrow is in its own component. Not sure if this is useful, but the code was already there. It will take up 100% of its container by default.
4. **BeamBtn:** Another basic wrapper, this time for buttons. Saves having to type an HTML button element and apply the beam_btn class each time. 
5. **BeamProgress:** A text element that has a progress label displaying a complete/incomplete message. This was basically the "status" element from the metadata component. The label and complete/incomplete messages can be set via props. 

I added a Components story that showcases these components. There is a "Text Elements" variant that uses mostly CSS modifiers applied to text elements (beam--alert, beam_copy, beam--bold). 

This is mostly a list of stuff I picked out from the metadata component. Let me know if there are any other types of components that could be useful.


agritheory:
@crabinak I know I mentioned that I want to turn this comment of yours into documentation, but I think @lauty95 needs these components to move forward.

PR #172, opened 2024-10-22
[node_editor] fix: add edge create, remove

### LINKED ISSUE #168: [node editor] Repair edge connections

robproject:
### Description
Add edge creation and removal handling to NodeEditor.vue

### Testing
Observe dragging a connection from one node to another creates an edge. 
Observe right-clicking the edge removes it.

### DIFF
--- a/common/changes/@stonecrop/node-editor/168_edge_connections_2024-10-22-01-17.json
+++ b/common/changes/@stonecrop/node-editor/168_edge_connections_2024-10-22-01-17.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@stonecrop/node-editor",
+      "comment": "Add handling for edge creation and removal.",
+      "type": "none"
+    }
+  ],
+  "packageName": "@stonecrop/node-editor"
+}
--- a/common/changes/@stonecrop/node-editor/168_edge_connections_2024-10-22-01-17.json
+++ b/common/changes/@stonecrop/node-editor/168_edge_connections_2024-10-22-01-17.json
@@ -3,7 +3,7 @@
     {
       "packageName": "@stonecrop/node-editor",
       "comment": "Add handling for edge creation and removal.",
-      "type": "none"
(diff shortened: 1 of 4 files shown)


agritheory (changes requested):
@robproject A couple of small things here to make the UX on this a little shinier:

- [x] Handle the ESC as "discard changes / same as clicking outside
- [x] Set default label width to 40ch when editing
- [x] Update Builder example with some edges

- [x] I think this text is probably unintended:
image

robproject:
>     * [x]  Handle the ESC as "discard changes / same as clicking outside
Looks like it already does this?

node_editor/src/components/StateEditor.vue
```diff
@@ -40,31 +40,26 @@ const elements = computed<FlowElements>({
...
-								source: key,
-								label: edgeKey,
-								animated: true,
-							})
-
-							// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
-							hasInputs[target] = true
-						}
-					}
+					const target = edgeValue.target || edgeValue
```
robproject, line 50: If the proxy array value.on has more than one edge, edgeValue will contain a proxy object where `target` can be accessed. Otherwise, edgeValue will be available directly.

agritheory: Great comment, let's put it in the code

common/changes/@stonecrop/node-editor/168_edge_connections_2024-10-22-01-17.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@stonecrop/node-editor",
+      "comment": "Add handling for edge creation and removal.",
+      "type": "none"
```
agritheory: Change to "patch" per @Alchez

agritheory (changes requested):

agritheory (approved):

PR #173, opened 2024-10-25
[atable] allow arbitrary HTML in table cells

Alchez:
Possible solve for https://github.com/agritheory/autoreader/pull/7.

---

**Concerns:**
- I've used Vue's `v-html` directive to render arbitrary HTML strings, but they mention [possible security concerns](https://vuejs.org/guide/essentials/template-syntax#raw-html) from using it, especially if it can render user-provided input.
- Taking the above point further, I'm not sure about the UX for possibly editing cells that have HTML elements as the cell value.

One possible solution that may solve both the concerns is to automatically disallow editing a cell if it's value is HTML, but that feels too restrictive and may possibly be solved with another modal?

What do you think?

### DIFF
--- a/atable/src/components/ACell.vue
+++ b/atable/src/components/ACell.vue
@@ -20,8 +20,8 @@
 			v-if="column.cellComponent"
 			:is="column.cellComponent"
 			:value="displayValue"
-			v-bind="column.cellComponentProps">
-		</component>
+			v-bind="column.cellComponentProps" />
+		<div v-else-if="isHtmlValue" v-html="displayValue" />
 		<span v-else>{{ displayValue }}</span>
 	</td>
 </template>
@@ -33,6 +33,7 @@ import { computed, CSSProperties, inject, ref, useTemplateRef } from 'vue'
 
 import TableDataStore from '.'
 import type { CellContext } from '@/types'
+import { isHtmlString } from '@/utils'
 
 const {
 	colIndex,
@@ -85,6 +86,14 @@ const displayValue = computed(() => {
 	}
 
 	return cellData
+})
+
+const isHtmlValue = computed(() => {
+	if (typeof displayValue.value === 'string') {
+		return isHtmlString(displayValue.value)
+	}
+
+	return false
 })
 
 const handleInput = () => {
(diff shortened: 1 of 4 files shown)


agritheory:
It is my understanding (though perhaps it is outdated, that Vue's `<component is="div">` should just work and that these extra step aren't strictly necessary.

Alchez:
> It is my understanding (though perhaps it is outdated, that Vue's `<component is="div">` should just work and that these extra step aren't strictly necessary.

@agritheory I tried that as well, and the `is` parameter can indeed take a tag name and render it. But there's a couple of problems with that:
- **Render HTML string:** If we use a `<div>` and pass the formatted HTML string as is, Vue will simply render it as plaintext. So an `<a>` tag wouldn't render in the DOM unless we use `v-html`.
- **Render HTML DOM element**: If we instead parse the string, extract the `<a>` tag into an Element variable and pass it to the `<component>`, that could work but I wasn't able to figure out how to arbitrarily pass DOMElement attributes to the component's `v-bind` function.

atable/src/components/ACell.vue
```diff
@@ -20,8 +20,8 @@
 			v-if="column.cellComponent"
 			:is="column.cellComponent"
 			:value="displayValue"
-			v-bind="column.cellComponentProps">
-		</component>
+			v-bind="column.cellComponentProps" />
+		<div v-else-if="isHtmlValue" v-html="displayValue" />
```
agritheory: Is there a way to do this without adding in this extra `div`?

atable/src/utils.ts
```diff
@@ -0,0 +1,5 @@
+export const isHtmlString = (htmlString: string) => {
+	const $temp = document.createElement('div')
```
agritheory: Document won't always be available in a Nuxt-like context, let's use DOMParser instead.

https://stackoverflow.com/a/15458968

agritheory (changes requested):

atable/src/components/ACell.vue
Alchez: Doesn't seem like there's an easy alternative, and someone had [asked for one](https://github.com/vuejs/vue/issues/7431) too. Generally you can use the `<template>` tag to avoid adding arbitrary elements, but that tag doesn't work with `v-html`. I could change it to `<span>` to avoid adding block elements though.

agritheory (approved):
Looks good! Thanks for figuring these quirks out

PR #177, opened 2024-10-28
[beam] add toast notification component

### LINKED ISSUE #164: [beam] Toast component
I imagine this to be a different component than the BeamModal one, but I could be convinced. 

In order to harmonize with the ActionFooter component, we want these to appear at the top/center of the interface.

crabinak:
@agritheory I added the toast notifications using [vue-toast-notifications](https://www.npmjs.com/package/vue-toast-notification). I was able to override styling in the beam style sheet to align the colors with the Beam-defined ones. There are some controls on Histoire that will let you change a few settings in case you want to see it in a different position. 

Let me know if this will work.


package.json
Alchez, line 1: @crabinak Don't use root-level `package.json` files, since Rush handles dependencies differently. You can either run the package install command inside the package directory directly or use [`rush add`](https://rushjs.io/pages/commands/rush_add/) instead.

common/changes/@stonecrop/beam/feat-beam-toast-notify_2024-10-28-18-12.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@stonecrop/beam",
+      "comment": "added toast notifications",
+      "type": "none"
```
Alchez, line 6: @crabinak This isn't documented anywhere, but I think you'll have to manually change this to `"type": "patch"` for generating changelogs

examples/beam/default.story.vue
```diff
@@ -62,6 +104,26 @@ const workOrder = reactive({
...
+const showNotification = () => {
+	let toastOptions = {
+		message: toastMsg.value,
+		position: toastPosition.value,
+		duration: toastTime.value * 1000,
+		type: toastType.value,
+	}
+	// let notification = toast.success("Notified!", toastOptions)
+	let notification = toast.open(toastOptions)
+}
```
Alchez: @crabinak @agritheory Should we create a wrapper around `vue-toast-notification` so that users can import it from `@stonecrop/beam` rather than having to remember to use the third-party library? Maybe we package the custom styles into our composable too?

Alchez (changes requested):

examples/beam/default.story.vue
agritheory: @Alchez @crabinak I agree. Curt is working on the style overrides but I don't think they're included here. There's a likely case where we want to use a toast notification in a desktop/aform type environment too,

Alchez: @agritheory @crabinak

Okay, so I tried to create a composable wrapper around the dependency but there seems to be an issue with it that I can't locate the source of.

- If I keep the dependency in `examples`, then I get errors during the build but toasts still work inside the story so the errors are non-blocking.
  - Someone already raised an [issue](https://github.com/ankurk91/vue-toast-notification/issues/97) last month for the errors I saw but the author isn't interested in maintaining the project anymore. 
- If I move the dependency to `beam` and try to create a re-usable composable, then the build fails entirely and I'm unable to access any of the Beam stories. I'm not sure what's going on there but this is blocking.

I think we'll have to either:
- look at another dependency
- somehow fix the existing dependency
- or recommend using the library manually in the target app (which kinda defeats the purpose of this PR)

What do you guys think?

Issue #175, opened 2024-10-27
[beam] Add Filtering options to Beam ListView


agritheory:
Header / slot for component to filter the list view or call API

Between one and three filtering options, each should be either a Boolean or select to facilitate positive mobile UX

PR #179, opened 2024-11-04
[beam] add list filter component

### LINKED ISSUE #194: [mobile] Scanning API fixes
- In a Stock Entry or Repack context when an item is scanned, the `add_or_increment` function does not call the appropriate `details` API to get the rest of the information.
image
- the `route` and filter` functions need to call the API to refresh the list.

crabinak:
@agritheory 
Here is a Filter component put together for the Beam ListView #175 . I added a FixedTop wrapper component for the NavBar and Filter to sit in to stick them both to the top without having to worry about height management. 

I set up the BeamFilter component to use BeamFilterOption sub components where key:value pairs can be passed in as props for each selection option. This component can basically be duplicated for each filter option you want. I opted to not use the HTML select element here as there is very little control over the styling of options. 

I tried to make the menus slightly larger to accommodate a user navigating with a finger on a mobile device.


beam/src/components/BeamFilter.vue
```diff
@@ -0,0 +1,71 @@
...
+let headerHeight = ref(null)
+
+const toggle = () => {
+	isOpen.value = !isOpen.value
+}
+
+onMounted(() => {
+	headerHeight = getTotalHeight(header.value)
+	totalHeight = getTotalHeight(beamFilters.value)
+	beamFilters.value.style.height = headerHeight
```
Alchez: The `ref` variables will need to be get/set reactively:

```suggestion
	headerHeight.value = getTotalHeight(header.value)
	totalHeight.value = getTotalHeight(beamFilters.value)
	beamFilters.value.style.height = headerHeight.value
```

beam/src/components/BeamFilterOption.vue
```diff
@@ -0,0 +1,129 @@
...
+		type: [Array, { choice: String, value: String }],
+	},
+})
+
+let value = ref(props.choices[0].value)
+let choice = ref(props.choices[0].choice)
+
+const updateValue = choiceData => {
+	choice = choiceData.choice
+	value = choiceData.value
```
Alchez: Same as above:

```suggestion
	choice.value = choiceData.choice
	value.value = choiceData.value
```

beam/src/components/BeamFilterOption.vue
```diff
@@ -0,0 +1,129 @@
...
+
+const props = defineProps({
+	title: {
+		type: String,
+		default: 'title',
+	},
+	choices: {
+		type: [Array, { choice: String, value: String }],
+	},
+})
```
Alchez: This is mostly syntactical sugar, but this could be rewritten as:

```suggestion
const { title = 'title', choices = [] } = defineProps<{
	title?: string
	choices: { choice: string; value: string }[]
}>()
```

This would also require the `<script>` tag to change to `<script setup lang="ts">`

beam/src/components/ToggleArrow.vue
```diff
@@ -0,0 +1,33 @@
+<template>
+	<div class="arrow-icon" :class="{ open: open }">
+		<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 35.36 70.71">
+			<polygon points="0 70.71 0 0 35.36 35.36 0 70.71" />
+		</svg>
+	</div>
+</template>
+<script setup lang="ts">
+import { ref, defineProps } from 'vue'
```
Alchez: Small nitpick: the `defineProps` macro doesn't need to be imported

common/changes/@stonecrop/beam/feat-beam-list-filter-style_2024-11-04-03-12.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@stonecrop/beam",
+      "comment": "added BeamFilter component",
+      "type": "none"
```
Alchez: I don't have a good suggestion to automate this, but this should be changed to the following for generating changelogs correctly.

```suggestion
      "type": "patch"
```

Alchez (changes requested):

crabinak:
@Alchez Thanks. I committed those changes to the branch.

Alchez (approved):

agritheory (approved):

PR #180, opened 2024-11-04
[beam] add DayDivider to list views

crabinak:
@agritheory 
I've added a DayDivider to the Beam ListView. I centered-aligned the text to help it visually stand out amongst the lists. The medium gray worked best here as the darker colors overpowered the navigation menu. This takes in month, day, and year as props and formats them to display, but it could be reworked to take in an object or however the date would be passed in.


examples/beam/default.story.vue
```diff
@@ -49,6 +49,28 @@
...
+			</BeamModal>
+
+			<Navbar @click="handlePrimaryAction">
+				<template #title>
+					<BeamHeading>Items to Receive</BeamHeading>
+				</template>
+				<template #navbaraction>Done</template>
+			</Navbar>
+
+			<BeamDayDivider :month="'10'" :day="'20'" :year="'2024'" />
```
agritheory: Something like this but perhaps with a pre-formatted date. 

```const itemsWithDayDivider = computed(() => {
	items.splice(5, 0, {
			label: new Date(),
			linkComponent: 'DayDivider',
		})
	return items
})```

agritheory (changes requested):
I'd like to see the specimen use the ListView component API, I left an example that works but  doesn't look right

image

Alchez:
@crabinak I've added a commit that merges changes from the main branch, so you will have to pull before making more changes.

crabinak:
@agritheory I updated this to use linkComponent. It accepts a Date object as the prop as well as an option to specify the format. As an example, I provided two date dividers: one as the default DateString and the other using ISOString. I didn't add more format types just incase this feature was not desired. 

I also left-aligned the DayeDivider content as having it center-aligned made it stand out but not in a good way.

Alchez:
@agritheory @crabinak 

I added some minor processing changes, and here's what the latest iteration looks like:

image

Let me know if that looks okay.

agritheory:
@crabinak @Alchez How do we feel about using the lighter blue "cell change color" as background for this?

crabinak:
@Alchez Looks good to me. 

@agritheory I updated this to use that light blue color you requested. I also removed the top border from the day divider as it was causing a double border with the bottom border from the previous item.

Alchez (approved):

PR #136, opened 2024-06-26
[project] Add API docs for each project

Alchez:
Changelist:
- Update .gitignore to latest version suggested by Rush 
- Add [Heft](https://heft.rushstack.io/) tooling to the Stonecrop monorepo
  - Includes a Typescript plugin to run for all projects
  - Includes [API Extractor](https://api-extractor.com/pages/overview/intro/) for auto-generating API structure
  - Includes [API Documenter](https://api-extractor.com/pages/setup/generating_docs/) to convert auto-generated API into documentation
- Move package stories into the `examples` folder to avoid dependency issues (eg. aform <-> atable circular dependency)
- Added ESLint fixes

Todo:
- [ ] Add package documentation

PR #181, opened 2024-11-05
[beam] add MQTT stream composable

### LINKED ISSUE #176: [beam] MQTT input stream component
Background: https://developer.mozilla.org/en-US/docs/Web/API/Streams_API

MQTT dependency TBD

This should probably be a composable. Similar VueUse composable: https://vueuse.org/core/useUserMedia/#useusermedia

Alchez:
Closes #176 

---

I tested the composable with an Artemis MQTT broker running locally (via https://github.com/agritheory/mqtt-pg-logger) and was able to see logging of published topic messages.

https://github.com/user-attachments/assets/fb0ded9f-b450-4762-84c2-bc6771aac3f7

---

**Notes:**
- The output of the composable is a list of messages grouped by the topic, but let me know if you want the signature to be different.
- Also considered using https://github.com/kaandesu/vue-paho-mqtt since it injects the MQTT client into the Vue instance directly, which is easier to interface, but it doesn't seem to support MQTT v5 yet. I'm not sure if that's a deal-breaker, but https://www.npmjs.com/package/mqtt atleast supports all protocol versions.
- There's another project that uses a similar composable pattern at https://github.com/tommy44458/mqtt-vue-hook using the `mqtt` library but I'm not sure if we want to package our own implementation or use that one.


agritheory:
> The output of the composable is a list of messages grouped by the topic, but let me know if you want the signature to be different.

The composable should ideally accept an array of topics to subscribe to and ignore all others. The topics would probably be managed as part of the application state (Pinia). If you want to subscribe to "all", use the "#" symbol, which is convention in MQTT.

Alchez:
@agritheory I've added a fallback to subscribe to all topics via the broker, in case a list of topics isn't provided to the composable.

common/reviews/api/beam.api.md
agritheory, line 1: What is this filing doing? I don't remember seeing a reviews folder before

agritheory (approved):

common/reviews/api/beam.api.md
Alchez: This is an API report generated by the [API Extractor plugin](https://api-extractor.com/pages/overview/intro/) that I had added back in #136.

It basically lets you verify the public API of the package before it gets converted into documentation by the [API Documenter plugin](https://api-extractor.com/pages/setup/generating_docs/).

Ref: https://api-extractor.com/pages/overview/demo_api_report/

PR #187, opened 2024-11-06
[atable] [aform] change cell style on value update

### LINKED ISSUE #186: [ATable][AForm] Add Table level configuration for colorization on cell-level value change
Might be good to have the component climb the schema to see if the parent has a configuration, so this could be enabled globally.

Alchez:
Closes #186.

---

@agritheory The feature described in the ticket kinda already existed, but it wasn't reactive so a cell change wouldn't actually trigger the styles to change. This PR simply makes the styles reactive now.

I'm not sure about the intended colorization API here though, and I see a couple of options (doesn't have to be either-or):
- Keep the status quo, and have the target application (`fab`, etc.) define the necessary custom CSS property (`--cell-modified-color`) in it's CSS stylesheets (or Vue style blocks, scoped or otherwise) at the `:root` level.
  - It seems like the Riverence app already defines this property in its stylesheet, so maybe this change fixes a regression?
- Add an additional key in the table config to define the cell-modified styles, which would allow individual tables (even if there are multiple in the same Vue file) to define it's own colors.

---

Here's some other things to consider that may or may not be out of scope of this PR:
- Currently, only the cell's background color is reactive. Should we consider other style properties like text color?
- Should the applied styles be global for all cells, or should each column define it's own cell-modified schema? Maybe it accepts a function that decides the styles based on the value?

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -9,14 +9,15 @@ jobs:
   changes:
     name: Check for changes
     runs-on: ubuntu-latest
+
     permissions:
       pull-requests: read
+
     outputs:
-      aform: ${{ steps.filter.outputs.aform }}
-      atable: ${{ steps.filter.outputs.atable }}
-      beam: ${{ steps.filter.outputs.beam }}
+      packages: ${{ steps.filter.outputs.changes }}
+
     steps:
-    - uses: dorny/paths-filter@v2
+    - uses: dorny/paths-filter@v3
       id: filter
       with:
         filters: |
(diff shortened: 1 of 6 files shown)


agritheory (changes requested):
@Alchez This should be implemented in one of the examples

Alchez:
@agritheory I've added the `--cell-modified-color` property to the default ATable stories. Here's how it works:

https://github.com/user-attachments/assets/a554e933-b44d-4243-b24b-54f6607f08cb

PR #191, opened 2024-11-15
[project] fix stonecrop styles

crabinak:
I have been updating the styling here and scoping most styles into components. I found some rogue generic class names floating about that I was able to scope or remove. There are now significantly less global styles now. I was also able to make some fixes to things like the expandable list in Atable not rendering correctly. I think I was able to track down that weird "table cell different sizes" glitch as well and hopefully remedy it. I was also able to update cells to properly ellipsis delimit content.  

I have run into an issue trying to run `rushx dev/beam` where I am now given the error: 

> 'The request URL [file/path/here] is outside of Vite serving allow list'

. Did something change that I need to update? This started happening after I pulled down the most recent changes to the development branch. Aform and Atable work fine. I was still able to at least move Beam styles into proper components for now.


Alchez:
@crabinak I've added a commit that should try and mitigate the filesystem errors, but since I couldn't replicate the issue locally, I'm not sure if it's a real fix. Can you pull my changes and try the Beam stories again?

agritheory:
@crabinak Are you anticipating any more work on this? I'd like to include it in an update.

crabinak:
@agritheory It should be good for now. I don't think there is anything major outstanding on this.

PR #196, opened 2024-11-27
[atable] allow resizable columns

### LINKED ISSUE #194: [atable] Make resizable columns a column-schema change

Alchez:
Closes #194

---

**Changes:**
- Allow table columns to be individually resizable using a new `resizable` key
  - If a column is marked as resizable, then the appropriate styles are automatically applied to the column's header cell
- Added a resize observer on all header cells that are resizable
- Remove min-width settings from all cells (this isn't exactly compatible with the resize observer since users wouldn't be able to reduce a cell's width after expanding it)
  - @agritheory How should the `full-width` setting work when columns are resizable? Status quo was to apply the `width: auto;` style, but since we need to set a default `width`, I'm not sure if that setting can be used anymore.
- Updated tests to reflect min-width removals

**Demo:**

https://github.com/user-attachments/assets/6ac4099d-97af-4c01-bec1-6b5dd119b607


agritheory:
> How should the full-width setting work when columns are resizable? Status quo was to apply the width: auto; style, but since we need to set a default width, I'm not sure if that setting can be used anymore.

I think the solution to this is to disallow resizing of the final column when it is configured as full-width.

agritheory:
We should hide the horizontal scrollbar on resizeable column headers:

image

Alchez:
@agritheory I've changed the `overflow` value from `auto` -> `hidden` for resizable columns, so the scrollbar shouldn't appear on the header cells anymore.

agritheory:
@Alchez I tested this and am still seeing the column creep to left

Alchez:
@agritheory I unfortunately still can't replicate this one locally, so it's been a hard one to debug :sweat_smile: 

I can't remember: does changing the `ch` values to `px` in the stories get rid of the creep in all instances of the table render? Just to verify if this is a `ch` -> `px` read/write issue from the resizable observer.

Alchez:
Implemented in #252 instead.

PR #201, opened 2024-12-02
[atable] Fix atable gray border

crabinak:
This adds a gray left border to Atable. I added this to the first cell of each row so that it doesn't cause issues with pinned column sizing. It works with both indexed and non-indexed rows since it just grabs the first td in the row.

I also added a variable `--sc-cell-changed-color` that uses the light blue color for cell changes. There was a stray `background:yellow` override in the default histoire file that I removed. 

Finally, vertical padding is reduced on rows.


agritheory:
@crabinak This color between the aform and atable left borders seems to be inconsistent. And I'm not sure that the expanded row in the expansion table shouldn't have a left border.  
image

crabinak:
@agritheory I updated the gray to match the aform border. One issue that arises is there are now two types of gray borders on atable: the table row dividers which are darker gray and the lighter gray of the left-hand border. 

atable-borders

I can think of two solutions:

1. Update the color of atable row dividers to match the light gray aform border color. 

atable-border-light

2. Update aform borders to match the darker gray of the atable row dividers. 

atable-border-dark


I'd probably be leaning towards using the darker gray for aform to match atable since the fields of aform are already darker, but let me know what you think.

crabinak:
@agritheory The atable now uses dark gray for its borders. I have a half size border for the expanded row (2px vs. 4px). I also updated aform's border color to match the atable border.

agritheory (approved):

Issue #200, opened 2024-12-01
[stonecrop] Implement SSR Support for Stonecrop


agritheory:
Stonecrop uses a singleton pattern to maintain a single source of truth for schema, registry and application state. The current implementation isn't SSR-compatible, causing errors when trying to access `window` during server-side rendering.

- [x] Maintain singleton pattern for schema/registry to preserve single source of truth
- [ ] Make mutable state (Pinia) SSR-aware
- [ ] Add SSR tests in playground environment

agritheory:
@Alchez I think we have the first of the three tasks done here, let's specify the other two more completely.

PR #204, opened 2024-12-03
[nuxt] add nuxt module

### LINKED ISSUE #199: [project] Create Nuxt Module
- [x] Create installable Nuxt module (@stonecrop/nuxt)
- [x] Integrate with Nuxt's routing system and use Stonecrop's router implementation for doctypes, throw error at build when conflicting routes exist
- [x] Provide configuration API in nuxt.config.ts if required - reserve router and docbuilder keys for now
- [x] Setup basic module structure with playground
- [x] Runtime plugin for Stonecrop initialization
- [x] Doctype auto-import should be enabled by globbing a @/doctype directory or specified "domains"- analogous to Frappe's modules/ "module def" 
- ~~Add schema-driven component auto-import - unused components should not be imported~~
- [ ] Add module option for docbuilder UI - this should be configurable as a module argument in nuxt.config.ts and could be sensitive to process.env for example
- [ ] Integrate event hooks from Workflow with [Nuxt lifecycle hooks](https://nuxt.com/docs/guide/going-further/hooks) - requires #169
- [x] AForm, ATable, any other custom components should be automatically registered

Alchez:
Closes #199.


agritheory:
This likely requires resolving #200 first

agritheory:
If you're looking for a shortcut for User schema, [this is what's minimally required](https://github.com/agritheory/mqtt-pg-logger/blob/3171a2e0c33674340cc4f61cb8c328bc17a97d58/src/gql.py#L145) for a PG-backed JWT auth workflow

And [here's a schema](https://gist.github.com/agritheory/520923fb93b13a8dd0550c4682127cc7) for authentication and authorization in Postgres

nuxt/src/runtime/plugin.ts
```diff
@@ -1,9 +1,11 @@
+import { AForm } from '@stonecrop/aform'
 import { Stonecrop } from '@stonecrop/stonecrop'
 import { useRouter } from 'nuxt/app'
 
 import { defineNuxtPlugin } from '#app'
 
 export default defineNuxtPlugin(_nuxtApp => {
 	const app = _nuxtApp.vueApp
+	app.component('AForm', AForm)
```
agritheory: This should probably use the plugin pattern to make all components available rather adding the components one-by-one.

agritheory:
I can see the Users UI briefly before it shows an error in the browser: "500 crypto.randomUUID is not a function" and a differetn one in the console:

```
 ERROR  [nitro] [unhandledRejection] Converting circular structure to JSON
    --> starting at object with constructor 'Object'
    |     property '_context' -> object with constructor 'Object'
    --- property 'app' closes the circle

  --> starting at object with constructor 'Object'
  |     property '_context' -> object with constructor 'Object'
  --- property 'app' closes the circle
  at JSON.stringify (<anonymous>)
  at Object.toDisplayString (/home/tyler/stonecrop/node_modules/@vue/shared/dist/shared.cjs.js:463:154)
  at Module.ssrInterpolate (@vue/server-renderer:194:35)
  at _sfc_ssrRender (src/runtime/layouts/StonecropHome.vue:26:100)
  at renderComponentSubTree (@vue/server-renderer:715:9)
  at renderComponentVNode (@vue/server-renderer:664:12)
  at renderVNode (@vue/server-renderer:779:14)
  at renderComponentSubTree (@vue/server-renderer:730:7)
  at renderComponentVNode (@vue/server-renderer:664:12)
  at renderVNode (@vue/server-renderer:779:14)


 WARN  [nuxt] Failed to stringify dev server logs. Received DevalueError: Cannot stringify arbitrary non-POJOs. You can define your own reducer/reviver for rich types following the instructions in https://nuxt.com/docs/api/composables/use-nuxt-app#payload.
```

Alchez:
@agritheory 

> I can see the Users UI briefly before it shows an error in the browser: "500 crypto.randomUUID is not a function"

This showed up in the FAB app before as well when using the ATable package, and I had helped Heather through the error [here](https://github.com/agritheory/fab/pull/46#issuecomment-2475004413).

If we assume secure contexts across the board, then I think this is a non-issue, but if we want to ease development as well in potentially insecure contexts, then I think we should revert to using the `uuid` module. What do you think?

> `ERROR  [nitro] [unhandledRejection] Converting circular structure to JSON`
> `WARN  [nuxt] Failed to stringify dev server logs. Received DevalueError: Cannot stringify arbitrary non-POJOs. You can define your own reducer/reviver for rich types following the instructions in https://nuxt.com/docs/api/composables/use-nuxt-app#payload.`

These errors/warnings are coming from trying to render the Stonecrop instance object as is into the page (`<pre>{{ stonecrop }}</pre>`), and are safe to ignore since we'll be removing that code eventually.

agritheory:
@Alchez Turns out this is a [known WSL problem](https://stackoverflow.com/a/70004971). I was trying to access it via the WSL IP which the browser has every reason to think is an insecure website. It also only works on Brave and not Firefox. That said, I think the limitations of using the browser for this are clear. Before refactoring the UUID package I think it might make sense to look at alternatives that are more performant.

Alchez:
@agritheory It seems like if we're looking at performance, then we might want to consider [`nanoid`](https://github.com/ai/nanoid). There's a bunch of performance benchmarks [(1)](https://prabeshthapa.medium.com/optimizing-your-system-with-the-right-unique-id-uuid-ulid-or-nanoid-78bf8b7bf200) [(2)](https://medium.com/@matynelawani/uuid-vs-crypto-randomuuid-vs-nanoid-313e18144d8c) and NanoID seems to be consistently fast while generating IDs (although slower than the native `crypto` module). Interestingly, `nanoid` themselves are saying they're - on average - about [half as performant](https://github.com/ai/nanoid#benchmark) than `uuid`.

The differences in performance also only seem to show up when you scale up the number of IDs being generated, which is currently low since they're only being used for table stores.

Should I move to `nanoid` or back to `uuid`?

agritheory:
I think nanoid is still overbuilt for our purposes and that a randomly generated 8-character base64 hash would be fine.

Alchez:
@agritheory I've added the `router` and `docbuilder` keys to the module options schema. Although I'm not sure how the docbuilder should unpack into the application; should it be an extra set of pages that re-implement the `docbuilder` example? Or should it expose the docbuilder components only?

agritheory:
> should it be an extra set of pages that re-implement the docbuilder example

Yeah, I think that's what's most helpful, but having the components available wouldn't hurt either

PR #212, opened 2024-12-05
[atable] added background gradient to rows with changed content in atable

crabinak:
@agritheory I added the light blue gradient effects to rows that have changed cell content. 

One note is that on pinned columns, there isn't a way to have a row's background override a cell's background with its own gradient (pinned cells need a background color to hide non-pinned content scrolling below it), so for that situation, I have the pinned cells use a solid light blue background color while the rest of the row continues the gradient as normal.

@Alchez I had to modify how the classes are applied to the cell in Acell since it now conditionally checks for pinned and cellModified, so I moved that logic into a computed property. Let me know if that causes any issues. 

image


agritheory:
@crabinak I think it's a little bit too much on the left and not enough on the right. I tried experimenting with this briefly enough to know I'm out of my depth, but found [a potential solution on SO](https://stackoverflow.com/a/71098929)

crabinak:
@agritheory I added some control in the style to adjust the lightness of the start and end colors for the gradient. I think the start value for the cell-changed blue is now 80% and the end is 40% (the values in the CSS are percentages of white added).

crabinak:
@agritheory Latest update here, the changed cells now have a black border, bold and italic text (might be too much), and the background of the cell is a slightly darker version of the cell-changed color.

agritheory:
@crabinak The border on the changed cell is too much I think. And I think that you were right before about the italics + bold and no background color on cell as the best way to communicate the data has changed. I would like to preserve the cell color change API, it might be a better UX in some cases.  CC @Alchez

crabinak:
@agritheory Should the background gradient be removed or kept?

agritheory:
@crabinak I like it, but I don't think it makes sense in all cases. When you're in a scenario where there a lot of columns / horizontal scrolling I think it's a good choice; we have a good use case for this and will implement it there.

crabinak:
@agritheory I updated so now modified cells just become italic and bolded. I included separate classes for cell highlighting (cell-modified-highlight) and row highlighting (changed-row-gradient) that can be applied for the cell and row highlighting where needed.

PR #226, opened 2024-12-18
[atable] Loading Options

### LINKED ISSUE #183: [ATable] Add loading skeleton and slot
At the table config layer we want to be able to provide a loading message and a slot for a customized loading element.

There are two cases we're running into currently: 1) table loading data from server and 2) table loading from state management. It's possible that the state management API is quite slow, it would be good to do some testing and eliminate the need for this.

crabinak:
@agritheory Here are two options for the ATable loading we discussed utilizing a background gradient animation. I have one using the gradient and another using a bar that moves across the bottom. This is a Loading component that could be used in place of the ATable before the data loads. We can explore other options as well. 

Setting this as a draft until we settle on a loading option as I have two similar components in there and unused ones would need to be cleaned up before merging.


agritheory:
@crabinak Both of these look great. I added a timeout on an additional sample to see how it could render in the table. I've never used the `suspense` component before, but it feels like that might be better than `v-show`.

Alchez:
> I've never used the `suspense` component before, but it feels like that might be better than `v-show`.

@agritheory I've never used it before either, but it seems like it's meant for component trees that have multiple async components, and it allows managing loading states across all of them in a better way.

Alchez (approved):

Issue #215, opened 2024-12-06
[beam] Add SegmentedDisplay component


agritheory:
To be used as a scale output. Numeric only, props to define number of integer and decimal places (perhaps use the mask functionality we've already implemented, and that as a prop) and the color as any valid css color (rgb, hex, name, variable, etc)

PR #228, opened 2024-12-20
[beam] Segmented Display Component

crabinak:
@agritheory I've created a Segmented Display component based on #215 . You should be able to change the CSS color of both the background and display text using any CSS terminology (accepts string names, variable names, RGB, RGBA, HSL, HSLA, and hex). Variable names should just be _--variable-name-only_ and not include the var() wrapper since the component handles that. 

This component takes the following props: 

- display-color (custom Color type looks for Hex, RGB/RGBA, HSL/HSLA, CSS Variable, or CSS string name)
- text-color (custom Color type looks for Hex, RGB/RGBA, HSL/HSLA, CSS Variable, or CSS string name)
- display-input (number)
- decimal-places (number)


agritheory (approved):
This came out perfect. Thank you!

PR #235, opened 2025-01-10
[atable] fix for modal positioning

crabinak:
@Alchez @agritheory related to the issue we were having with the amodal not positioning correctly. This fix changes the modal position to _fixed_ while still auto-detecting edges (so it doesn't get cropped off-screen). The position of the modal uses the position of the cell instead of offset positioning, so it shouldn't matter what the parent is doing (I used some test wrapping divs as parents to ATable and gave them various positioning). 

Let me know if this resolves that weird positioning issue that was occurring.

Example image: The modal right aligns with the page since it would otherwise overflow and get cropped. 
fix-modal-position


Alchez:
@crabinak I think the logic for rendering modals for the last few cells has maybe regressed. Here's what I see in the ATable default story

image

image

crabinak:
@Alchez I added a line to include the Y position of the window scroll when calculating the maximum height. Let me know if this fixes it now.

Alchez (approved):
@crabinak This looks good. But I've added a commit based on some typing warnings in the IDE. Can you please check out the change locally and verify that the behaviour hasn't changed?

crabinak:
@Alchez It is still working fine. Thanks!

PR #236, opened 2025-01-10
[aform] add adropdown features

lauty95:
- Handle external clicks: Modified the handleClickOutside function to ensure results only close when clicking outside the component.
- Fixed result selection: Updated the setResult function to pass the selected result correctly to closeResults.
- Improved result closing: Adjusted the closeResults function to handle optional parameters and only clear the search when necessary.


aform/src/components/form/ADropdown.vue
```diff
@@ -1,8 +1,8 @@
 <template>
-	<div class="autocomplete" :class="{ isOpen: isOpen }">
+	<div ref="autocomplete" class="autocomplete" :class="{ isOpen: isOpen }">
 		<div class="input-wrapper">
 			<input
-				ref="mopInput"
+				ref="adropdown"
```
Alchez: I don't think we need this ref yet

aform/src/components/form/ADropdown.vue
```diff
@@ -29,86 +29,84 @@
...
+const { label, items, isAsync, filterFunction } = defineProps<{
 	label: string
 	items?: string[]
 	isAsync?: boolean
+	filterFunction?: (search: string) => Promise<string[]>
 }>()
-
-const emit = defineEmits(['filterChanged'])
-
+const autocomplete = ref<HTMLElement | null>(null)
```
Alchez: This can use the inbuilt `useTemplateRef` Vue macro instead

aform/src/components/form/ADropdown.vue
```diff
@@ -29,86 +29,84 @@
...
-	} else {
-		results.value = items.filter(item => {
-			return item.toLowerCase().indexOf(search.value.toLowerCase()) > -1
-		})
+const handleClickOutside = (event: MouseEvent) => {
+	if (autocomplete.value && !autocomplete.value.contains(event.target as Node)) {
+		closeResults()
+		arrowCounter.value = 0
 	}
 }
```
Alchez: I think we can revamp the click-outside logic using the [VueUse directive](https://vueuse.org/core/onClickOutside/#directive-usage)

aform/src/components/form/ADropdown.vue
```diff
@@ -29,86 +29,84 @@
...
 	isOpen.value = true
-	if (isAsync) {
+	if (isAsync && filterFunction) {
 		isLoading.value = true
-		emit('filterChanged', search.value)
+		try {
+			const filteredResults = await filterFunction(search.value)
+			results.value = filteredResults
+		} catch (error) {
+			console.error('Error en el filtrado asíncrono:', error)
```
Alchez: What kind(s) of error are we expecting here?

examples/aform/dropdown.story.vue
```diff
@@ -32,21 +32,24 @@ const async_dropdown_data = reactive({
...
 })
 
-function filterItems(search: string) {
-	setTimeout(() => {
-		async_dropdown_data.items = async_dropdown_data.allItems.filter(item => {
-			return item.toLowerCase().indexOf(search.toLowerCase()) > -1
-		})
-	}, 750)
+function filterItems(search: string): Promise<string[]> {
+	return new Promise(resolve => {
```
Alchez: You don't need to manually return a promise, setting the function to be `async function` will automatically setup a Promise instead.

Alchez (changes requested):
@lauty95 I think the dropdown should return all possible results when clicked, but I'm only seeing the selected value (the test failure might be related too).

aform/src/components/form/ADropdown.vue
lauty95: Sounds great! I'll implement it.

examples/aform/dropdown.story.vue
lauty95: It's true, but I had to create "delay" function to return a promise, otherwise it doesn't work

aform/src/components/form/ADropdown.vue
lauty95: I'm not sure, but if the `filterFunction` fails for any reason, I thought of catching it. However, I added a return of an empty array.

lauty95:
> @lauty95 I think the dropdown should return all possible results when clicked, but I'm only seeing the selected value (the test failure might be related too).

I changed the onFocus event to display all possible results when clicked.
I also updated some tests because I changed the way asynchronous functions are called.
However, I still see the tests failing due to this error: "ERROR: Coverage for branches (66.66%) does not meet the global threshold (70%)".

examples/aform/dropdown.story.vue
Alchez: That's weird, what's the issue/error if you don't use a delay function?

aform/src/components/form/ADropdown.vue
Alchez: Oh that's probably good to do, and we can later think about propagating the error to the UI too.

aform/src/components/form/ADropdown.vue
```diff
@@ -29,86 +28,89 @@
...
-import { onMounted, onUnmounted, ref /* useTemplateRef */ } from 'vue'
-
-const { label, items, isAsync } = defineProps<{
+import { ref, useTemplateRef } from 'vue'
+import { vOnClickOutside } from '@vueuse/components'
+const { label, items, isAsync, filterFunction } = defineProps<{
 	label: string
 	items?: string[]
 	isAsync?: boolean
+	filterFunction?: (search: string) => Promise<string[]>
```
Alchez: A couple of things:
- Should the `filterFunction` be allowed to be sync as well? The filter function you've added in the test is really just a sync function, so maybe we support both?
- Should the items also be passed to the `filterFunction`? So that the user has context on the current list of items.

aform/src/components/form/ADropdown.vue
```diff
@@ -29,86 +28,89 @@
...
 }>()
-
-const emit = defineEmits(['filterChanged'])
-
+const autocomplete = useTemplateRef<HTMLElement>('autocomplete')
```
Alchez: You're not using the ref anywhere, I think you should remove this.

PS: for safety, avoid naming the ref variable the same as that in the template since it seems like that causes an [issue](https://github.com/vuejs/core/issues/11795#issuecomment-2326858438) setting other ref values.

aform/src/components/form/ADropdown.vue
```diff
@@ -29,86 +28,89 @@
...
 const results = ref(items)
 const search = defineModel<string>()
 const isLoading = ref(false)
 const arrowCounter = ref(0)
```
Alchez: I think we should rename `arrowCounter` to something more useful: maybe `activeItemIndex`?

aform/src/components/form/ADropdown.vue
```diff
@@ -29,86 +28,89 @@
...
-
-onUnmounted(() => {
-	document.removeEventListener('click', handleClickOutside)
-})
 
-const setResult = result => {
-	search.value = result
+const closeResultsHandler = () => {
 	closeResults()
+	arrowCounter.value = 0
```
Alchez: Should this be unset (`null` or `undefined`?) if there's no active item being selected anymore?

aform/src/components/form/ADropdown.vue
```diff
@@ -1,11 +1,10 @@
 <template>
-	<div class="autocomplete" :class="{ isOpen: isOpen }">
+	<div ref="autocomplete" class="autocomplete" :class="{ isOpen: isOpen }" v-on-click-outside="closeResultsHandler">
```
Alchez: I think the function name can be more instructive:

```suggestion
	<div ref="autocomplete" class="autocomplete" :class="{ isOpen: isOpen }" v-on-click-outside="closeDropdown">
```

Alchez (changes requested):
> However, I still see the tests failing due to this error: "ERROR: Coverage for branches (66.66%) does not meet the global threshold (70%)".

@lauty95 This is fine to ignore. As we add more tests, that number should go beyond the threshold.

examples/aform/dropdown.story.vue
lauty95: The truth is... doing this:

```javascript
async function filterItems(search: string): Promise<string[]> {
	const filtered = async_dropdown_data.allItems.filter(item => item.toLowerCase().includes(search.toLowerCase()))
	async_dropdown_data.items = filtered
	return filtered
}
```

works fine... however, you won’t see the "delay effect" that you would get if the function had some delay.
Does that make sense?

aform/src/components/form/ADropdown.vue
lauty95: Good to know, thanks!

examples/aform/dropdown.story.vue
Alchez: I think I'm confused, why do we need a delay effect at all? I would assume the dropdown filtering is performant and so it would have less visual "lag"

lauty95: I assumed that `ADropdown` has an `isAsync` prop to allow filtering elements by calling an API endpoint or something similar. The "delay effect" would simulate of waiting for a response from the server.

I also added a third example in dropdown.story.vue 
Screenshot 2025-01-17 at 11 36 12 AM

That example shows how to use a custom filter function

aform/src/components/form/ADropdown.vue
lauty95: I guess filterFunction could support both, async in case that could call an API or similar and sync if you need a local custom filtering.

Regarding passing `items` to the function, I don’t see a clear use case where it’s necessary. Do you think in some case?

Alchez: > Regarding passing items to the function, I don’t see a clear use case where it’s necessary. Do you think in some case?

Hmm, that's true too. I suppose if the context on the items is available in the component itself, then we don't need to worry about passing in the items.

examples/aform/dropdown.story.vue
Alchez: Oh, I see what you're going for. For some reason, I thought this piece of code was living inside the main component code, but doing this simulation inside the story is a good idea 👍

Sorry for the confusion!

Alchez (approved):

Issue #239, opened 2025-01-28
[atable] Table modals show above pinned columns


Alchez:
While implementing tables in FAB, I noticed that the modal renders above pinned columns, which creates a weird UX when scrolling inside the table.

I'm not sure what the expected behaviour should be:
- Should the modal close on scroll to avoid this kind of behaviour to exist?
- Should the modal have a lower z-index to hide behind pinned columns?
  - Basically the z-index priority tree would be: pinned columns > modal > table
  - I tried implementing this and it fixes the issue, but the modal is also still visible outside the container's bounds, which seems odd

PR #240, opened 2025-01-29
[atable] modal z-sorting

crabinak:
@Alchez @agritheory This is to hopefully address #239. 

In order to clip the modal within the table container, I had to change it from fixed to absolute positioning. I had to rework how it positions itself as a result. The modal now gets the cumulative width of previous cells within the row/height of previous rows in the table to position itself correctly. I added an extra "pinned_extra_columns" variant in the Histoire list story where you can see the modal display under the pinned columns (I also wrapped atable in a positioned div to make sure it is positioning relative to the table and not the page) . It feels like this works a bit better than before as it is now contained within the table instead of sort of floating above everything.

I set the pinned columns to z-index 10 and the modal is 5, so we can have some leeway in between if we need it.  

_There is one small glitch that is confounding me where the modal is visible within the 1px row border when it is beneath sticky columns. I have no idea what is causing this, but let me know if you are seeing it as well._ 
pinned-cols-glitch


agritheory:
@crabinak I'll bet that's a transparent outline or border instead of one that would match the table (inherited?) bg-color

Alchez:
@crabinak I changed the style logic to use offsetValues instead, which were resolving to the same value as the calculated ones. Can you check and let me know if they look fine to you as well?

> There is one small glitch that is confounding me where the modal is visible within the 1px row border when it is beneath sticky columns. I have no idea what is causing this, but let me know if you are seeing it as well.

I am seeing this as well, and I'm not sure what could be causing it. Claude seems to think that it has something to do with the table properties, but I'm not convinced.

crabinak:
@Alchez Thanks. For some reason I thought offsetValues gave issues last time, but that was probably when using a fixed modal. 

I found a fix for the border issue. Since the row's border was set on the row and not the table cells, the modal ended up z-sorted above the row with borders but below the sticky content. I moved border-top from ARow to ACell and it seems to have fixed it.

Alchez (approved):

PR #241, opened 2025-01-30
[beam] add emit response on beam filter option set

Alchez:
I've added an emit to Beam filter options that should help us react to filters being set via the UI. This contains work towards https://github.com/agritheory/beam/issues/210.


beam/src/components/ListView.vue
```diff
@@ -1,6 +1,6 @@
 <template>
 	<ul class="beam_list-view">
-		<li v-for="item in items" :key="item.label">
+		<li v-for="item in items" :key="item.barcode || item.label">
```
Alchez, line 3: @agritheory We're setting the `key` attribute to identify each item uniquely. The problem here is that the app doesn't expect `barcode` to always be set, in which case I've set the fallback to the item's `label`.

In case neither of these are unique, list filtering will be inaccurate since any items with duplicate keys will not be re-rendered when the filter is applied.

I'm not sure if hashing the object could be a possible solution, but what do you suggest?

beam/src/composables/mqtt.ts
Alchez, line 1: The changes in this file are irrelevant to the Beam filters, but because the MQTT server wasn't running, it would constantly error and stop the Histoire server as well. Adding a port check to make sure the service is running before we spawn the client.

Alchez (commented):
@agritheory There's a design concern here as well. If a Beam list view contains day-divider elements, and the user filters the data, should the day-dividers be applied again, or should they disappear to show the result?

Currently, we're adding the day-dividers manually to the Histoire stories, but they could be computed instead by the `ListView` component automatically, possibly behind a toggle.

PR #243, opened 2025-02-24
[beam] use two-column layout for beam filters

Alchez:
Updates layout for https://github.com/agritheory/beam/pull/260.

image


agritheory:
@Alchez @crabinak I think we need to have a flex vs grid discussion here - and I think I may have specified this wrong in the first place. 

If it's really narrow, I think we'd want cleaner stacking: 

image

And if it's really wide, I think we'd want some kind of max-with. 
image

Maybe we need an AForm `<select>` element? Unifying these design languages is an ongoing, unfinishable project, but this one seems like a good opportunity.

crabinak:
@Alchez @agritheory Filter options will now stack when the window is under about 500px. Additionally, each filter option has a max width of 300px, so at larger sizes, they won't stretch too large. I converted the Filter Options container to use flexbox instead of CSS Grid so filter options can flow from left to right instead of being contained within grid columns. Let me know if this remedies the issue.

agritheory:
@crabinak Can you set the break at something like 479? The target device is 480x854.

crabinak:
@agritheory Updated to 479px.

agritheory (approved):

Issue #242, opened 2025-01-30
Gantt view / overlay for atable


agritheory:
I think this can be a row type. Let's start with drag-able handles. There will be a lot of logic that will be implementation specific, so it may make sense to make this very bare-bones to start. How and where the Gantt-specific data is stored will be an important early decision, but should probably be in a new key - a peer to columns, rows and config. 

![Image](https://github.com/user-attachments/assets/537ef421-1560-4f55-b737-e63822be11a7)

Issue #245, opened 2025-03-01
Add `watch` flag or functionality to examples scripts


agritheory:
https://rushjs.io/pages/advanced/watch_mode/

Alchez:
@agritheory Some things to watch (heh) for after these changes: https://github.com/agritheory/stonecrop/pull/248#issuecomment-2700043757.

PR #248, opened 2025-03-01
ATable Gantt integration

agritheory:
#242 

- [x] Colspan API, it might make more sense to make this generalizable, for now its tied to the Gantt view type 
- [ ] Design API to accept color for handler from row data, can be assumed to be on the first column. Handles are currently a darker blue- they should be similar to the size they are now but thematically gray
-  ~~Design API for labeling the handler with ellipsis~~
- [ ] Design API to locate start and end for handler (it can be assumed that the columns are time periods of some kind)
- [x] CSS/ Layout issues:
    - [x] Gantt Rows are taller than other rows
    - [x] zIndex issue on the handles (image)
     - [x] Header columns are borked and I don't understand why

image

To be solved in future work in FAB or other implementations:
- A default modal component to directly edit the date/ time of the task/ handler data
- Emitters on drag / release (for example to re-allocate the units in the task)


agritheory:
@Alchez Can you figure out the header issue? 
@crabinak Can you tackle some of the styling issues?

crabinak:
@agritheory Sure thing. I've fixed the row height and handle-indexing issues for now.

Alchez:
@agritheory I've added a [commit](https://github.com/agritheory/stonecrop/pull/248/commits/18d9feb7a181b64c29e059bab8625be009d0fe1a) that should fix the header issue. It seems like the `<tr>` elements in the table were having their `display` attribute overridden from `table-row` (default) to `flex`, which was somehow causing the header columns to be pushed beyond the content. Reverting that seems to fix the styling issue.

@crabinak Can you please also check the commit and confirm the changes?

image

PS:
- I've added a merge commit that brings in the changes from #245, so running `rushx dev:atable` should automatically enable watch mode.
  - Note that adding a commit while watch mode is on will fail with the message: `Another Rush command is already running in this repository.`. Simply stopping the histoire server should let you proceed.
  - Maybe moving the Rush hooks from `pre-commit` to `pre-push` would be less friction?
- Tests are failing and our coverage went way down :sweat_smile:

agritheory:
@Alchez @crabinak Nice! 
Curt can you tackle this one: 
 - Design API to accept color for handler from row data, can be assumed to be on the first column. Handles are currently a darker blue- they should be similar to the size they are now but thematically gray

I will add color to the data. 
 
Rohan, can you pick up this one?
 - Design API to locate start and end for handler (it can be assumed that the columns are time periods of some kind)

crabinak:
@agritheory I added in some functionality to read the defined color and apply it to the bars. The handles and borders all use opacity so they adapt to the base color instead of needing to define separate shades. If no color is specified, the default #cccccc gray is applied.

agritheory:
@crabinak I think this styling where the line is just barely colored is too subtle.

image

crabinak:
@agritheory For some reason, your view is still showing the bars not filling 100% row height. I attached a screenshot here of how they should be looking. 
bars-screenshot

crabinak:
@agritheory, I was able to fix the Gantt bar height issue, so it should work correctly now in Firefox.

agritheory:
@Alchez Could you review and merge?

Alchez:
@agritheory The feature looks good. I added a commit that makes the resizing snappier.

PS: Our coverage has [dropped considerably](https://github.com/agritheory/stonecrop/pull/248#issuecomment-2699976678). Should we think about adding tests before merging this one?

PR #250, opened 2025-03-03
[beam] render filter menus over list content

Alchez:
Fixes https://github.com/agritheory/beam/pull/260.

---

@agritheory @crabinak I added some minor fixes to get the filters to overlay instead of moving the list content down:
- Set the filter dropdown menus to be `position: absolute;` so that they're outside the normal DOM flow and can be overlayed over other content.
  - However, doing the above makes the filter menus draw over the list content rather than pushing it down, but the menus then stay hidden behind the list content.
- Changed the `overflow` setting for the `beam_filters` container element from `hidden` to `visible` to allow the menus to actually render in front of the list content.
  - However, doing the above _always_ shows the dropdown menus instead of having them only visible behind the filter toggle.
- Modified the top-level options container in `BeamFilter.vue` to only show options content if the section is open.

The `overflow: visible;` setting may have other unintended side-effects as shown by the last change I had to make, but I'm not sure how else to handle this?

---

image

image

### DIFF
--- a/beam/package.json
+++ b/beam/package.json
@@ -41,6 +41,7 @@
 		"test:ui": "vitest --ui"
 	},
 	"dependencies": {
+		"@vueuse/components": "^12.7.0",
 		"@vueuse/core": "^12.7.0",
 		"mqtt": "^5.10.3",
 		"onscan.js": "^1.5.2",
--- a/beam/src/components/BeamFilter.vue
+++ b/beam/src/components/BeamFilter.vue
@@ -5,7 +5,7 @@
 			<BeamHeading>Filter</BeamHeading>
 		</div>
 
-		<div class="beam_filters-options">
+		<div v-if="isOpen" class="beam_filters-options">
 			<slot></slot>
 		</div>
 	</div>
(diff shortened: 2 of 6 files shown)


crabinak:
@Alchez This looks good to me. The only issue I can see arising from this is that one filter option menu will overlap the other, but this is solved by just closing all Filter Option menus when an option menu is opened. I'm not certain how to emit this from the Filter Option to its siblings (or to the parent Filter from Filter Option and then to the child Filter Options) since Filter uses slots instead of explicit components.

Alchez:
> @Alchez This looks good to me. The only issue I can see arising from this is that one filter option menu will overlap the other, but this is solved by just closing all Filter Option menus when an option menu is opened. I'm not certain how to emit this from the Filter Option to its siblings (or to the parent Filter from Filter Option and then to the child Filter Options) since Filter uses slots instead of explicit components.

@crabinak I tried using the `v-on-click-outside` directive to manage each menu's local state in itself, so they're not concerned with other menus. Can you review again and let me know what you think?

crabinak:
@Alchez Looks good! I added a few small style edits. One adds a gap between the filter options, the other left-aligns the options as they looked really strange floating in the center, particularly at larger sizes.

agritheory (approved):

PR #252, opened 2025-03-11
[atable] id to ref

### LINKED ISSUE #247: Resizable should be refactored to another query handler
We will have multiple ATables on a page and that means the ID will resize all ATables. 

```vue
// https://github.com/agritheory/stonecrop/blob/development/atable/src/components/ATableHeader.vue#L2

<template>
	<thead id="resizable" v-if="columns.length">
		<tr class="atable-header-row" tabindex="-1">

```

lauty95:


atable/src/components/ATableHeader.vue
```diff
@@ -24,12 +24,21 @@
...
+const resizable = ref<HTMLElement | null>(null)
+
 const { columns, store } = defineProps<{
 	columns: TableColumn[]
 	store: ReturnType<typeof createTableStore>
 }>()
+
+onMounted(() => {
+	if (resizable.value) {
+		console.log('ATableHeader mounted:', resizable.value)
```
agritheory: Remove log?

agritheory (changes requested):

agritheory:
@lauty95 I think we talked about the `useTemplateRef` composable?

lauty95:
@agritheory @Alchez 

The id="resizable" and ref usage were removed because column resizing is handled directly by modifying the width property inside the reactive columns array.

Since arrays in JavaScript are passed by reference, when multiple ATable instances share the same columns array, resizing one column will affect all tables.

To avoid this and ensure each table has its own reference to the columns:

```js
const default_table = reactive({
	rows,
	columns: [...columns], // ← This creates a new reference
	config: { view: 'list' },
})
```

IRL usage, this may not always be necessary, as columns will often come from different data sources.

## Preview

eg

Alchez:
@agritheory @lauty95 there was a previous attempt to implement resizing over at #196. Can we bring in anything from there that would help?

lauty95:
@Alchez Thanks for sharing that PR with me, I didn't see it! I'll review it to see what we can incorporate. I really like the effect you achieved there.
Do you know why it ended up staying as a draft? Was there something specific that needed more work? Maybe we could revive #196?

Let me know what you prefer: should we continue with #196 or incorporate some ideas from that PR into this one?

Alchez:
@lauty95 I think it probably just went down in priority, but we can discuss with @agritheory on when this is needed.

For the change itself, ideally we want the best parts from both the PRs. Can you see if there's any gaps between the two implementations and maybe merge them into this PR?

lauty95:
@Alchez Sure! I'll analyze both PRs

lauty95:
@Alchez I made a kind of merge from #196 

I used `v-resize-observer` instead vanilla JS.

Let me know if it's ok or need some change.

examples/atable/default.story.vue
```diff
@@ -2,6 +2,8 @@
 	<Story title="default">
 		<Variant title="default">
 			<ATable v-model="default_table.rows" :columns="default_table.columns" :config="default_table.config" />
+			<ATable v-model="default_table_2.rows" :columns="default_table_2.columns" :config="default_table_2.config" />
+			<ATable v-model="default_table_3.rows" :columns="default_table_3.columns" :config="default_table_3.config" />
```
Alchez: I think these should go into their own variant; the `default` story variannt should just show a simple table.

examples/atable/default.story.vue
```diff
@@ -111,7 +114,19 @@ const readonly_columns: TableColumn[] = [
...
-	columns,
+	columns: [...columns],
+	config: { view: 'list' },
+})
+
+const default_table_2 = reactive({
+	rows,
+	columns: [...columns],
+	config: { view: 'list' },
+})
```
Alchez: There's some performance issues going on in rendering the story now. A visible lag before it's usable. I'm not sure what's causing it, but here's a warning from trying to parse the column data: [atable-default-performance.log](https://github.com/user-attachments/files/20116445/atable-default-performance.log)

Alchez (changes requested):
@lauty95 I've added a couple of small comments about the ATable stories, but the functionality has been merged well

examples/atable/default.story.vue
lauty95: A little bug :sweat_smile:

https://github.com/user-attachments/assets/e0724de0-204d-4ccd-8898-392769aca0a4

There was a bug on `onResize` function. I just fixed that.

As for the warning, I noticed it's also present in the development branch (could you confirm that on your end?).

I tried to debug it, and it seems to come from this line:
https://github.com/agritheory/stonecrop/blob/b1b99a80358a05a8eebb50eb93957fd7963fda81/examples/atable/default.story.vue#L57

From what I understand, the issue happens because functions inside the props can't be cloned properly. If I rewrite the code like this:

```js
<Variant title="default">
	<ATable v-model="default_table.rows" :columns="columns" :config="default_table.config" />
</Variant>

const columns: TableColumn[] = [...]

const default_table = shallowReactive({
	rows,
	config: { view: 'list' },
})
```

Then the warning doesn't appear, unless you reload the page.

Let me know how you'd like to proceed with the warning, and if the latest changes are ok

Alchez: @lauty95 I think we can ignore the warning since it seems to be coming specifically from Histoire trying to parse the ATable stories (opening the story in a new tab doesn't show that warning). The other changes look good too :+1:

Alchez (approved):

lauty95:
@agritheory this is ready to be merged

PR #253, opened 2025-03-14
Sticky Column outline bug fix

### LINKED ISSUE #246: sticky column offset issue
I think we might need to add some margin or padding - it's not clear to me what the root issue is:

Image

crabinak:
@agritheory I couldn't quite replicate #246 but I added a 1px left margin to the cells. Let me know if this alleviates the overlap issue.


Alchez:
@agritheory I can't replicate this locally either. Is this still required?

agritheory (approved):

PR #257, opened 2025-04-15
[atable] handle gantt emit events

Alchez:
Work towards https://github.com/agritheory/fab/issues/90.


Alchez:
@agritheory I think we'll need to talk through the design for this one. Here's some general notes:

- The component structure for gantt rows looks like this: `ATable` > `ARow` > `AGanttCell`, which means that the emit must be caught and handled by the `ARow` implementation.
  - This may not always be ideal if a user does not want to modify the existing `ARow` implementation, since they'll be forced to re-implement it to handle the emits.
  - We could work around this in two ways:
    - By bubbling up the emit right up to the `ATable` component so that the base implementation can handle it.
    - (Recommended) We could use the existing Pinia store for the table, and run actions on each emit event. This way, we can subscribe to those events outside of the `template` tags as well.
- The render structure for gantt rows looks like this: `1 Gantt Row > N Data Rows`.
  - The above only works visually, since there is no common binding between the gantt and data rows in the component or table store.
  - We need to implement this binding to be able to correctly react to the emit events via the UI. This also needs to consider arbitrary fields like `resource_group` as in the story example for ATable-Gantt.

I've made some changes that expects a `gantt` key in the main gantt row, with details like color, start and end indexes, and optionally a colspan. But I'm wondering if that `gantt` key could be re-used for data rows as well, and it could define a group binding there as well?

agritheory:
I think this will work just fine!

PR #261, opened 2025-06-13
[beam] ItemCount validation

lauty95:
While working on a Beam issue, I noticed that in some`ListView` inputs, clicking on them would display a "NaN" value.

Also, when the "denominator" value isn't required, it's not possible to update the input value.

After checking some Stonecrops files, I found that the ItemCount.vue component had a validation rule that was causing these issues. I modified the validations, and now everything works as expected.

## After

https://github.com/user-attachments/assets/90aedbfb-d397-4c57-ba15-e1738eab2814

## Before

https://github.com/user-attachments/assets/9740935d-e7bb-434f-b2e5-5f7e80eac70a

### DIFF
--- a/beam/src/components/ItemCount.vue
+++ b/beam/src/components/ItemCount.vue
@@ -15,7 +15,7 @@
 
 <script setup lang="ts">
 import { useDebounceFn } from '@vueuse/core'
-import { computed, nextTick, type HTMLAttributes } from 'vue'
+import { computed, type HTMLAttributes } from 'vue'
 
 const count = defineModel<number>({ required: true })
 const {
@@ -33,15 +33,8 @@ const {
 const isCountComplete = computed(() => denominator !== 0 && count.value === denominator)
 
 const validate = (payload: ClipboardEvent | InputEvent | MouseEvent) => {
-	const newValue = Number((payload.target as HTMLElement).innerHTML)
-	if (typeof newValue !== 'number' || isNaN(newValue)) {
-		count.value = count.value || 0
-	} else {
-		count.value = denominator > 0 ? Math.min(newValue, denominator) : newValue
-	}
-	nextTick(() => {
-		;(payload.target as HTMLElement).innerHTML = count.value.toString() || '0'
-	})
+	const newValue = Number((payload.target as HTMLElement).innerHTML) || 0
+	count.value = denominator ? Math.min(newValue, denominator) : newValue
 }
 
 const debouncedRequest = useDebounceFn((payload: InputEvent) => validate(payload), debounce)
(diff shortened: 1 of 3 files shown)


beam/src/components/ItemCount.vue
```diff
@@ -33,8 +33,20 @@ const {
...
+			;(payload.target as HTMLElement).innerHTML = '0'
+		})
+		return
+	}
 	count.value = Math.min(newValue, denominator)
+	if (denominator > 0) {
+		count.value = Math.min(newValue, denominator)
+	} else {
+		count.value = newValue
+	}
```
Alchez: Since the `NaN` is coming from the `Math.min` operation, I think this can be simplified to just:

```suggestion
	const newValue = Number((payload.target as HTMLElement).innerHTML) || 0
	count.value = denominator ? Math.min(newValue, denominator) : newValue
```

beam/src/components/ItemCount.vue
```diff
@@ -15,7 +15,7 @@
 
 <script setup lang="ts">
 import { useDebounceFn } from '@vueuse/core'
-import { computed, type HTMLAttributes } from 'vue'
+import { computed, nextTick, type HTMLAttributes } from 'vue'
 
 const count = defineModel<number>({ required: true })
 const {
```
Alchez, line 21: I think we'll also have to make the `denominator` prop optional with or without a default. But we should think about class colors in that case. Should a count without a denominator be tagged with the `beam--alert` class?

CC: @agritheory

Alchez (changes requested):

beam/src/components/ItemCount.vue
lauty95: @Alchez I tested your suggestion, but it doesn't handle all edge cases, such as when alphabetical characters are entered.

However, I made some improvements to the validation logic:

```ts
const validate = (payload: ClipboardEvent | InputEvent | MouseEvent) => {
	const newValue = Number((payload.target as HTMLElement).innerHTML)
	if (typeof newValue !== 'number' || isNaN(newValue)) {
		count.value = count.value || 0
	} else {
		count.value = denominator > 0 ? Math.min(newValue, denominator) : newValue
	}
	nextTick(() => {
		;(payload.target as HTMLElement).innerHTML = count.value.toString() || '0'
	})
}
```

demo:

https://github.com/user-attachments/assets/08276b34-946c-4d60-8d1f-cd45e5fb53a6

What do you think?

beam/src/components/ItemCount.vue
lauty95: @Alchez 
I made `denominator` optional but added a default value of `0`.

Regarding the `beam--alert` class, I agree that it shouldn't be applied when there's no valid denominator to compare with. To fix that, I updated the `isCountComplete` logic.

beam/src/components/ItemCount.vue
Alchez: @lauty95 Hmm, is there a use-case where alphabetical characters should be allowed?

lauty95: I don’t think so — I can't think of any use case.
@agritheory , what do you think? Should we allow alphabetical characters in this Vue component?

lauty95:
@agritheory Is this ready to be merged?
Once it's merged I can update dependencies in Beam, right?

Alchez (approved):

Alchez:
@lauty95 you should be able to update the dependency version in Beam now

PR #268, opened 2025-07-02
[atable] add gantt-tree base + story

### LINKED ISSUE #262: [atable] Combine Gantt row with Tree table views
- [x] Allow tree view to show Gantt cells
- [x] Design indent level / dependency data APIs

### LINKED ISSUE #265: [atable] Add node handle to AGanttCell

### LINKED ISSUE #266: [atable] create bezier path between AGanttCell nodes
- <strike>Use path components from VueFlow</strike>
- [x] Support all combinations of left-to-left, right-to-right, right-to-left and left-to-right
- [x] Support many-to-many but not duplicate paths between nodes

Alchez:
Closes #262 

[Screencast From 2025-07-02 16-48-30.webm](https://github.com/user-attachments/assets/82bba860-4c97-4936-bb2b-172a0d8870e2)


Alchez:
@agritheory This definitely needs some polish, but what do you think generally about the view?

agritheory:
@Alchez This is a great start. The columns widths are being weird in the video but the expansion behavior is what we're looking for. 
Rohan and @crabinak:
How do we feel about reducing the height of AGanttCell to ~80% of the line height?

Alchez:
@agritheory can you test this again? I've resolved the shikiJS errors you were seeing as well, and added a bunch of tests for gantt

agritheory:
@Alchez In Histoires full screen, the column snapping is not working and is pretty messed up. Still missing the bottom scroll bar too.

image

crabinak:
@agritheory @Alchez Small change, I changed the pointer-events on connections-svg from auto to none. The SVG was positioned over half of the table, preventing the grabbing of handles beneath it.

Alchez:
@agritheory @crabinak I'm not sure what's happening exactly with the column widths in the `tree` story: all the period columns are set to be `10ch`. The snapping should be related to that issue too.

agritheory:
Screenshot 2025-07-14 at 08-45-12

@crabinak Design direction required:
- should the handles of the nodes remain visible as start and end?
- we can move the arrow icon or ">" icons to the middle of the bezier path. We'll want to accommodate a scenario where the paths are inverse of each other: A -> B; B -> A
- perhaps a dashed line to limit the amount of interference that the sold line has when it runs through a data cell?

crabinak:
@agritheory I'll take a look at what styling options vue-flow offers. I think having the node handles appear on hover works pretty well (currently how they work) instead of having them always visible, so it doesn't pollute the chart with extra information. The node handles are only relevant when hovering over a bar to connect it anyway. 

I do know dashed line is an option, so I can switch that. I might also try out a "smooth step" instead of bezier curve for these connections, since the gantt content is rigid while bezier seems more suited to a loose layout (like a drag and drop whiteboard).

agritheory:
@crabinak All good ideas, go ahead and chase it. Thanks! Rohan discovered that the vue-flow Bezier curve component wasn't really doing much so we implemented our own wrapper around a `path` element. Styling options are wide open.

crabinak:
@agritheory @Alchez I added the arrow to the middle of the path. The way marker-mid works with SVG is a bit weird in that it adds a marker at every vertex, not the middle of the path, so I needed to implement a workaround to place the arrow there. Alternatively, we could try and break the bezier into two paths, forcing a vertex in the middle, but I was not sure how to accurately calculate half a bezier curve. 

I noticed the curves only update upon dragging/moving a bar when that bar is hovered over. I wasn't sure what was triggering the refresh, but the arrows for some reason do not refresh. I'll keep looking at splitting the curve as I believe that will more accurately pair the arrow marker with the drawn curve.

Alchez:
@agritheory @crabinak I've resolved the merge conflicts and changed the ATable API (instead of accepting rows via `modelValue` and `rows`, which seems confusing, I've changed it so that only accepts a required `modelValue` value).

> I noticed the curves only update upon dragging/moving a bar when that bar is hovered over. I wasn't sure what was triggering the refresh, but the arrows for some reason do not refresh.

I'm not sure what you're seeing. I can see the arrows move after a bar has stopped dragging/resizing. Can you share a video?

crabinak:
@Alchez The arrows refresh with the lines now since I've switched them over to use markers, but the issue still remains where a curve will only update when the bar is connected to is moused over.

Alchez:
@agritheory @crabinak I'm gonna merge this one so that tree-views can be used in other projects. But let's put the pending improvements into a new PR.

PR #283, opened 2025-07-26
[stonecrop] Hierarchical State Tree

### LINKED ISSUE #278: [stonecrop] HST Design
Implement the core Hierarchical State Tree (HST) interface with unified API compatibility across Vue reactives, Pinia stores, and immutable objects.

- [x] Create `HSTNode` interface with path-based addressing (`get()`, `set()`, `has()`)
- [x] Implement `HSTProxy` class with JavaScript Proxy wrapper
- [x] Support dot-notation paths (e.g., `"users.123.profile.settings"`)
- [x] Maintain hierarchical navigation (parent/child/root relationships, breadcrumbs, depth tracking)
- [x] Handle Vue reactive, Pinia store, and immutable object access patterns
- [x] Integrate with global Stonecrop Registry singleton (no passing required)
- [x] Automatically define/provision path /emit address from Registry on instantiation of doctype from schema
- [x] Add watcher to `useStonecrop` composable to collect component-level mutation bubble up

agritheory:
#278


Alchez:
@agritheory I think this one is in a good enough spot now, can you have a look again?

agritheory:
OK some hacking on this and I have some review notes that we need to tackle:

- [ ] Remove hardcoded route parameters like (`records`/`record`) from the composable - these shouldn't be hardcoded in framework code. The route _is_ a request for data, as if it were a REST API; this is because it's a strong user expectation not because it's an implicit decision about a backend. Use a key for the PK pk the document instead.
- [x] Eliminate the doctype hierarchy configuration - The hierarchy object in `server.ts` should be computed from the registry. This is a new function. This won't be a part of the framework but represents a first step on a schema validation feature. 
- [ ] If a doctype contains nested doctypes, express this through the schema using composition syntax (e.g., `{'...SalesOrderItemList'}` within the sales-order schema), not through explicit parent/child key value pairs. This is a new feature.
- [x] Routes should only be registered after a successful meta endpoint call returns the doctype definition. Don't pre-configure route patterns on app start, for example, lazy load on request.
- [x] One meta endpoint; the schema IS the configuration - everything else should be computed from it.
- [ ] Doctype fragment type to load partials in a summary view.

Alchez:
@agritheory Let's discuss these over the call today, and I'm also adding a couple questions that we can answer there or async here:

> The route is a request for data, as if it were a REST API; this is because it's a strong user expectation not because it's an implicit decision about a backend

- Should the mapping between a route and it's doctype ever be present in the application itself? Or should that always be fetched from the server?
  - If the latter is true, should we simply pass the route to the server and expect it to resolve it to a doctype (or set of composed doctypes) and return their meta(s)?
  - Or should we first resolve the route in the application using a defined map (either in the app or fetched from the server) before retrieving its meta?

> Routes should only be registered after a successful meta endpoint call returns the doctype definition. Don't pre-configure route patterns on app start, for example, lazy load on request.

Are you referring to the hierarchy pre-registering routes for all related doctypes, or the catch-all route pattern in the desktop example that resolves a route to a doctype?

---

I'm also linking a [sequence diagram](https://mermaid.live/edit#pako:eNrVWVlv20YQ_isLPtQ2IMm1ZUcy0SRI5Tgx4ASBY-eh8MuKHIlbUyS7u5TjGP7vmT24vHRQPQy0QREe8-3O8c18S-XJC9IQPN8T8FcOSQDnjM45XdwlBP_LKJcsYBlNJLkVwNtPr9NcAidUkG852Lvfpvzwzf67LItZQCVLk4M2bpIusjQBvLJQ96AzWtBpDAqeC_gqERvwNDPoC4wAHlJ-vwJ7DXMmJH9UyOJ6O8ptoGDlzWUiJMWsbV_g49cbBVV_IZx3QHwFvjSpNVeH775crkqOQX5OJZBUAVShehbjk4ot-T1NJfmFfOHpnxBIcoWxk890yeb6tVlHoftv3phK-sVrIDIlh5kBCmNpTNDWlc43z0gQ0WSOEM7mc-CCxGqjoLAyaAcqFtDl9GvF3D8gfcINwxZIU4Nsxluieyq9PnFJxfIwyWjMfpj4ZynXQVc80DgVr2WCTyYRBPfa1O7MbI1t1NYQMUWKP7y_IYc0Y4cLkPRtmAbyMYPXNln92G1o7Pu13a5B5jzRyYkBIxJBBAtKHpiMSMSAUx5Ej5WIMaWRJOnM1fcaBOZRALoZxHkIwtccqe7eI8UdRrXo6feSinv7Ul-Wb6gQbJ4sVGXM-8oDZdVKQxkMDcNzEz1WTmWQxrGNSKB_hFtf20tYtplHYIA6_YJQDNdlv56DShq1rV9StFdeHvosNKEdqlDVK_135XkZonpbuVM2VdbVOON46pMEHsqpsM-tX7ap3QvEaII6WoKeB0LNg83rC6RJtm8jUhS2eT5YSY0m7gD5lczYPMcC6B0zKiNhYl-iHyFVibbUC2hGpyxGD8GaZBxwLqEF2lGSZshK1U5iezsWJK1NnHO1ylVKQ5bMN0c9B3kNQcpDsb9XJfTeisS2mnFT_1X2sA1orc2oUnGuKZ2e3cYKiany6JPqTgNuHDZwhBTACQc1RjUSuZEHuC2YPl_ksWTY_qQ-X1u5aYxa4zguG0i2xPLDTOBwV-Wd5UlQqVBr1BYLJCHWqxb5ksFDE4QoJQoYe5Q-1MxXE8Dojw5a1_sywY6m2iH074LFeOtKb-Wm4pYxINNHTBOVudgsF6g0YQy42URrzv6TKkmPIKtz5N-MQRwmqAbPByuYpl3EDqmTazDTDoiB2X5PraIefFNLHrTqepup7rE22mVoe9yF08U-YldyvzWOvjY82E51sw2EDbq1qG4j07xosbpJJUvC3GCc8nPoc02zzaRqurSVXY3jiaIWHlCKMXPhZKpNrwmehu5RUYjIIGAzFhR7tqu28QykL_pHx8NdD0NKQ_-NwxDZDyhO7JCENSnY4XSEyRFkYhZRKSNWVDYejz4AHuUMpnqqKLxoSHsjnuK0Y_BGcLrIa03-lKs7yh9gi0yRUxEIS-XVEji5vj1fIYAV9XTjdsoSJWB_QwRN---ogOWwUNnGYbFX8G8nMVTgBnO7qKLNmVYrfXZaMzLwALjV3Z6W1_YgNcraElWFL0R1UPe8BU6wzsgr7SGutGWRQSWSF9Za3TEum5Vpt011SyAsphCGRbRblVhdNpS44KOZ82LdwHwfMun2l0zG8JKKvKZyyo09u-haTdZijIMX4D_QrdtLxYJYSYE-_Vcys4tk3aj676ZXquAdxcp85-gPvJPTV-uV6jMG4CTJffA6hdGqc_CiUlUKzmq90nnbUa_ch-5LiJXa7B8qlfv-xwGR3Le1SH2UYR2CiMVYI4jNV1nEspau6doKpkmH7Yh16fL1pntJzbiP1hEW0Jh8wXYmn2hC57CoH2BqXWx-RilDqPnX6thPlCUS_6_HpDW6ZXuFyeg02gd14nfRaE2rukCbcTtJcXZ-l53V2nFNaV_hRyepdkjXt29trK-7C7eWFe145ccMgpPZTfKgGtC6s39ULbz7al0n3dWdcKi4QFx5ioAGFdNO-otStWQhuN8uStVd-UvZZhHWuSmFdGVC1unwGuwUTwZhwPPFVOwuAe9cMroJQZm8jipQ_T3LXPdH47P_kxpolt1EqOT9K1hC7CbSY4dJ9L4YqUTqFWK9QoMyFTZf5HHc5RDaGDJVUg-aae7I75nautZ0nVhZ7lw5VK5odq_nzTkLPR87GXreAviCqlvvSa1_50mUOLjzfLwMYUbzWN55d8kzwjKa_JGmiwKJ1ZxHnj-jscA7c3iy_3DkTHSzTdI8kZ5_Oh7rNTz_yfvu-f3RyXhwNj4Zj4fDs9Ho-KznPeLTk7PB6enx0a_Hp8ej4Wh0MnzueT_0rieD8dHRCf4Zvjoang1Hzz8BPPo2Sg) (you may need to disable dark mode for readibility) for how I'm imagining a user flow works in an example Stonecrop application. We can discuss this on the call today.

stonecrop/src/composable.ts
```diff
@@ -116,14 +98,41 @@ export function useStonecrop(options?: {
...
+							} catch (error) {
+								formData.value = initializeNewRecord(doctype)
+							}
+						}
+					} else {
+						formData.value = initializeNewRecord(doctype)
+					}
+
+					if (hstStore.value) {
+						setupDeepReactivity(doctype, recordId || 'new', formData, hstStore.value)
```
agritheory, line 277: Refactor to UUIDv7

PR #295, opened 2025-08-12
[stonecrop] add global watcher for v-model updates

Alchez:
Work towards #278


Alchez:
@agritheory can you verify the tests (`integration.spec.ts` and `working.spec.ts`) and documentation that Claude has generated here, and if it follows what you expected? That way, I can modify and cleanup the implementation accordingly.

I didn't want to pollute the main PR with too much auto-generated code, so I'm treating it as a feature branch for this specific change.

agritheory:
@Alchez I'm not sure I understand the long-term value of hst-reactive-failing.spec.ts, but I think it's fine for now. I'd like to find a  way of using real components in the tests instead of mocking them. And the completeness of the the test should include all the fieldtypes eventually, but this is the first pass, let's include a checklist to track it.

PR #298, opened 2025-09-11
[atable] Feat sorting and filtering

### LINKED ISSUE #258: [atable] sorting and filtering options
Placeholder, to be further defined

https://tristen.ca/tablesort/demo/

lauty95:
# Sorting and Filtering

https://github.com/agritheory/stonecrop/issues/258

## Demo

The demo video shows the implementation using example code in `examples/atable/default.story.vue`

https://github.com/user-attachments/assets/1c09e7b5-645c-4c68-835f-d3ee521a5c8c

Includes styling contributions from @crabinak in https://github.com/agritheory/stonecrop/issues/258#issuecomment-3344955873

## Changes

I renamed the ATableFilter.vue component to ATableColumnFilter.vue. This new component renders the individual column filters directly beneath their corresponding header titles in ATableHeader.vue.

Additionally, the column title in ATableHeader.vue now includes the handleSort function, which works when the column is sortable.


common/changes/@stonecrop/atable/feat-sorting_and_filtering_2025-09-11-13-00.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@stonecrop/atable",
+      "comment": "Sorting and Filtering feature added by props",
+      "type": "none"
```
Alchez: Let's change to "patch"


```suggestion
      "type": "patch"
```

atable/src/components/ATableColumnFilter.vue
```diff
@@ -0,0 +1,212 @@
...
+	column: TableColumn
+	colIndex: number
+	store: ReturnType<typeof createTableStore>
+}>()
+
+const filterValue = ref<any>('')
+const filterStartValue = ref<any>('')
+const filterEndValue = ref<any>('')
+
+const getFilterType = (column: TableColumn): string => column.filterType || 'text'
```
Alchez: Let's make this a `computed` property since it only needs to be resolved once per component trigger

atable/src/components/ATableColumnFilter.vue
```diff
@@ -0,0 +1,212 @@
...
+})
+
+// Filter actions
+const updateFilter = (type: string, value: any) => {
+	if (!value && type !== 'checkbox') {
+		store.clearFilter(colIndex)
+		filterValue.value = ''
+	} else {
+		filterValue.value = value
+		store.setFilter(colIndex, { type, value })
```
Alchez: I'm not sure where the `type` is being used for the filter, is this required?

atable/src/components/ATableColumnFilter.vue
```diff
@@ -0,0 +1,212 @@
...
+
+const { column, colIndex, store } = defineProps<{
+	column: TableColumn
+	colIndex: number
+	store: ReturnType<typeof createTableStore>
+}>()
+
+const filterValue = ref<any>('')
+const filterStartValue = ref<any>('')
+const filterEndValue = ref<any>('')
```
Alchez: Since these are related, I wonder if it's better to use a `reactive` instead?


```suggestion
const dateFilter = reactive({
	startValue: '' as string,
	endValue: '' as string,
})
```

Alchez (changes requested):
@lauty95 This looks great, but I think we should look into the following:

- [UX] The filter input boxes should maybe have a placeholder value of the column you're filtering (ex. Homepage)
- [UX] The date range filtering shouldn't start until the user has selected a date, because currently switching views to a different month/year automatically picks up a date and applies the filter
- [UI] I think the theming should maybe adjust for contrast, since I'm unable to see the clear filters/clear sort buttons on white background pages.

Image

PS: If possible, let's also add tests for these additions. And check the failing tests for verify changes.

CC: @agritheory

atable/src/components/ATableColumnFilter.vue
lauty95: I reviewed it. I'm not sure it's necessary to make that function a computed property since it's a simple function.
I removed it and now I'm just using `column.filterType` directly in the v-if instead of calling `getFilterType()`.

What do you think?

atable/src/components/ATableColumnFilter.vue
lauty95: Applied!

atable/src/components/ATableColumnFilter.vue
lauty95: Good question.
At the beginning I thought it might be necessary, but after reviewing it I agree that the type prop isn't needed, so I removed it.

lauty95:
> [UX] The filter input boxes should maybe have a placeholder value of the column you're filtering (ex. Homepage)

Each filter is displayed under its corresponding header, and I also based the design on Curtis's reference: https://github.com/agritheory/stonecrop/issues/258#issuecomment-3344955873

Are you sure we should include a placeholder?

> [UX] The date range filtering shouldn't start until the user has selected a date, because currently switching views to a different month/year automatically picks up a date and applies the filter

Absolutely agree. I changed the @input event on `<input type="date" />` to @change, which allows navigating between months or years without triggering the filter until a date is actually selected.

> [UI] I think the theming should maybe adjust for contrast, since I'm unable to see the clear filters/clear sort buttons on white background pages.

What about something like this:

Screenshot 2025-11-25 at 11 05 36 AM

Screenshot 2025-11-25 at 11 06 04 AM

atable/src/components/ATableColumnFilter.vue
Alchez: Yeah, that's even better 👍

Alchez:
@lauty95 

> > [UX] The filter input boxes should maybe have a placeholder value of the column you're filtering (ex. Homepage)
> 
> Each filter is displayed under its corresponding header, and I also based the design on Curtis's reference: [#258 (comment)](https://github.com/agritheory/stonecrop/issues/258#issuecomment-3344955873)
> 
> Are you sure we should include a placeholder?

That's a good point too. I think it's fine if we skip a placeholder for now.

> > [UI] I think the theming should maybe adjust for contrast, since I'm unable to see the clear filters/clear sort buttons on white background pages.
> 
> What about something like this:
> Screenshot 2025-11-25 at 11 05 36 AM Screenshot 2025-11-25 at 11 06 04 AM

Yeah, that looks great!

atable/src/stores/table.ts
```diff
@@ -186,6 +186,20 @@ export const createTableStore = (initData: {
...
+		const filterState = ref<
+			Record<
+				number,
+				{
+					value: any
+					startValue?: any
+					endValue?: any
+				}
+			>
+		>({})
```
Alchez: This seems very hard to read, can we extract this into a separate type instead?

common/changes/@stonecrop/aform/feat-sorting_and_filtering_2025-11-25-13-12.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@stonecrop/aform",
+      "comment": "Edited API.md to document new sorting and filtering features in ATable component.",
+      "type": "none"
```
Alchez: "patch" on these ones too

Alchez (changes requested):
I think this is pretty close. Let's also add tests, since our coverage has dropped dramatically 😅 

PS: I'm headed on vacation until Dec 2, so I won't be able to test this out by myself, but I could probably try give you more feedback via mobile.

lauty95:
Thanks for your review, @Alchez! I'll add some tests for these features.

Enjoy your vacation!

atable/src/stores/table.ts
lauty95: ready!

lauty95:
I added some tests to cover sorting and filtering

Also added some tests in `aform/tests/dropdown.spec.ts` and `aform/tests/form.spec.ts` to raise the coverage from 65% to 70% and avoid the following error:

```sh
Coverage for branches (65.46%) does not meet global threshold (70%)
```

Issue #206, opened 2024-12-04
[atable] Add tests to ensure that a table that is not editable isn't actually editable


agritheory:
We have a lot of internal APIs and I want to make sure that we capture every interaction we allow when the table is configured and editable, and not when it isn't.

PR #299, opened 2025-09-12
[atable] ACell tests - non-editable cases

lauty95:
https://github.com/agritheory/stonecrop/issues/206

**Non-editable behavior tests:**
1. **DOM attributes** - Validates `contenteditable="false"` and `data-editable="false"`
2. **Click interaction** - Confirms text selection is prevented on click
3. **Focus interaction** - Ensures `selectAllText()` is not called on focus
4. **Navigation state** - Verifies `currentData` updates for consistent navigation
5. **Input events (SKIPPED)** - Identifies potential bug where non-editable cells still process `@input` events

## Input events

The skipped test reveals that `ACell.vue` always has active `@input` listeners regardless of `column.edit` state. Non-editable cells can still emit update events.

Should I..
- Delete the skipped test
- Fix the component listener (add edit state validation)

## Results

Screenshot 2025-09-12 at 2 46 49 PM

**Any edge cases we should add?**

### DIFF
--- a/atable/src/components/ACell.vue
+++ b/atable/src/components/ACell.vue
@@ -250,6 +250,8 @@ const restoreCursorPosition = (position: number) => {
 }
 
 const updateCellData = (payload: Event) => {
+	if (!column.edit) return
+
 	const target = payload.target as HTMLTableCellElement
 	if (target.textContent === currentData.value) {
 		return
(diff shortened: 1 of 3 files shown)


Alchez:
@lauty95 I think we should look into the update emit cause that might create unexpected side-effects, rather than deleting the test. Good catch though!

PR #301, opened 2025-09-17
[stonecrop] cleanup stonecrop-hst-desktop example

Alchez:


examples/desktop/index.ts
```diff
@@ -35,25 +35,19 @@ const getMeta = async (doctype: string) => {
 const pinia = createPinia()
```
agritheory, line 48: What is this pinia instance being used for?

stonecrop/src/plugins/index.ts
```diff
@@ -1,9 +1,137 @@
...
+	registry: Registry,
+	stonecrop: Stonecrop,
+	doctype: string,
+	actualDoctype?: string,
+	apiBaseUrl = '/api'
+): Promise<void> {
+	try {
+		const targetDoctype = actualDoctype || doctype
+		// For list views, we need the list metadata which uses the '-list' suffix
+		const listDoctypeKey = targetDoctype.endsWith('-list') ? targetDoctype : `${targetDoctype}-list`
```
agritheory: I thought we were going to use a separate doctype for this instead of appending "list" to the doctype name

agritheory (commented):
Some convention-based questions

stonecrop/src/plugins/index.ts
Alchez: Yep, this is what I'm working on right now. Not sure why/how it got added into the plugin again.

examples/desktop/index.ts
Alchez: This is used for compatibility with setting up the ATable stores, but we could probably change that up later.

agritheory: Let's add a ticket for it. I think we want the HST to do that work instead, so removing this would be in scope of the HST feature.

PR #307, opened 2025-10-02
[graphql][stonecrop] Connection Fix

### LINKED ISSUE #264: [graphql][stonecrop] Load data and run actions from Stonecrop via GraphQL
- [ ] Integrate getRecord via query with arguments for doctype and id and an args object for future filtering implementation
- [ ] Integrate runAction via mutation with arguments for doctype, id and args object
- [ ] `getMeta`
 ---
Let's use the Mirage GraphQL plugin, already added to dependencies
We're going to need a way to add records to the HST without triggering the mutation/emits. Currently there isn't a separate API for this. I propose `add` or `addRecord` for the non-emitting version and `set` or `update` for the one where we want the changes to be tracked.

lauty95:
## WIP

- `getMeta` method added to Stonecrop Class
- General fixes to use Stonecrop Class from Graphql Client Example.
- `fetch` key added to allow Mirage connection.

### About fetch:

I encountered this error:

Screenshot 2025-10-02 at 10 58 40 AM

Adding the fetch key allowed Mirage to connect to `GraphQLClient`

## Next Step.

https://github.com/agritheory/stonecrop/issues/264

### DIFF
--- a/common/reviews/api/stonecrop.api.md
+++ b/common/reviews/api/stonecrop.api.md
@@ -282,6 +282,7 @@ export class Stonecrop {
     constructor(registry: Registry);
     addRecord(doctype: string | DoctypeMeta, recordId: string, recordData: any): void;
     clearRecords(doctype: string | DoctypeMeta): void;
+    getMeta(doctype: string): Promise<any>;
     getRecord(doctype: DoctypeMeta, recordId: string): Promise<void>;
     getRecordById(doctype: string | DoctypeMeta, recordId: string): HSTNode | undefined;
     getRecordIds(doctype: string | DoctypeMeta): string[];
--- a/examples/dev-runner.sh
+++ b/examples/dev-runner.sh
@@ -27,12 +27,17 @@ case $EXAMPLE_NAME in
     "desktop")
         PACKAGE_NAME="@stonecrop/desktop"
         ;;
+    "docbuilder")
+        ;;
+    "graphql_client")
+        PACKAGE_NAME="@stonecrop/graphql-client"
+        ;;
     "node_editor")
         PACKAGE_NAME="@stonecrop/node-editor"
         ;;
     *)
         echo "Unknown example: $EXAMPLE_NAME"
-        echo "Available examples: aform, atable, beam, code_editor, desktop, node_editor"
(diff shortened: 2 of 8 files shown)


lauty95:
@Alchez 
Rohan, I'd like your opinion about the problem I commented on and the solution I proposed.

The error occurs when Mirage.js creates the server and a POST method is called. The `getMeta` method from `graphql_client/src/index.ts` is executed, but the issue is with `client.request`.

I've left a test/debug in `graphql_client/src/index.ts`:

```js
const methods = {
    getMeta: async (doctype: string, url?: string): Promise<MetaResponse> => {
        const client = new GraphQLClient(url || '/graphql', {
            // fetch: window.fetch,
            jsonSerializer: {
                stringify: obj => JSON.stringify(obj), // process the request object before sending; leave as default JSON
                parse: metaParser, // process the response meta object
            },
        })
        console.log('Fetching meta for doctype:', doctype, 'from URL:', url || '/graphql')
        console.log(client)
        try {
            const { getMeta } = await client.request<Meta['response'], Meta['variables']>({
                document: queries.getMeta,
                variables: { doctype },
            })
            return getMeta
        } catch (err) {
            console.error('Error fetching meta')
            console.log(err)
            throw err
        }
    },
}
```

This throws the error: 'Uncaught (in promise) Ee: GraphQL Error (Code: 404)..'

Screenshot 2025-10-02 at 2 13 01 PM

The solution was to uncomment `fetch: window.fetch` which forces GraphQLClient to use the browser's native fetch API that Mirage.js can properly intercept. Without this, GraphQLClient may use its own HTTP implementation that bypasses Mirage's request interception.

Alchez:
@lauty95 your solutions seems good to me, especially since it doesn't seem like the folks over at Mirage are maintaining their [grapqhl library](https://github.com/miragejs/graphql) for a while now.

PS: I've also updated the script for running the GraphQL example: running `dev:graphql_client` inside the `examples` folder should also watch for changes in the graphql client package and its dependency packages, and automatically rebuild. Should ease up development without having to manually build for every change.

lauty95:
@Alchez Great! The updated script works well, thanks!

PR #308, opened 2025-10-07
[stonecrop] track action mutations and allow rolling back transaction if possible

### LINKED ISSUE #281: [stonecrop] Registration of actions on field-level signals
Create a system for registering and automatically triggering doctype actions when specific field paths are mutated, bridging browser-side mutations with the Stonecrop action system.

- [x] Design API for registering field-level action triggers in doctype definitions
- [x] Register all actions in registry with fully qualified paths, added in `addDoctype`
- [x] Automatic action triggering on HST `set()` operations pass path via dependency injection to function(s)
- [x] Support for multiple actions per field change
- [x] Action execution with change context (operation, before/after values, use dependency injection)
- [x] Each function should complete or error before the next one executes, probably needs a wrapper of some kind, with rollback
- [x] Error handling for missing or invalid actions, with rollback

These options aren't mutually exclusive but I do think [there should be one-- and preferably only one --obvious way to do it.](https://peps.python.org/pep-0020/#the-zen-of-python) I think we will want to use a pattern that favors composables. 

```
actions: {
  'emailAddress.*.is_primary': ['validateEmailPrimary', 'UPDATE'],
  'profile.name': ['(userInput) => {validateName(userInput)}'],
  'print': ['async () => { await globals.printUtils.getPrintFormatFromDialog()}']
}
```

Alchez:
Closes #281


stonecrop/docs/TRANSACTIONAL_TRIGGERS.md
```diff
@@ -0,0 +1,245 @@
...
+      profileId: profileData.id,
+      fieldsUpdated: ['email', 'lastLogin'],
+      previousValues: { email: context.beforeValue }
+    },
+    rollback: () => restoreProfile(profileData.id, context.beforeValue)
+  }
+})
+```
+
+### 3. **Automatic Rollback on Failure**
```
agritheory: I think a revert-to-set-point approach is easier to implement and more common in general. This is a lot of code for  something that that's not actually idempotent - there's not guarantee that rollbacks won't also fail.

stonecrop/docs/TRANSACTIONAL_TRIGGERS.md
```diff
@@ -0,0 +1,245 @@
...
+```
+
+## Usage Patterns
+
+### Basic Transactional Actions
+
+```typescript
+import { FieldTriggerEngine } from '@stonecrop/stonecrop'
+
+const engine = new FieldTriggerEngine()
```
agritheory: Maybe this should be a singleton pattern like registry?

stonecrop/docs/TRANSACTIONAL_TRIGGERS.md
```diff
@@ -0,0 +1,245 @@
...
+```typescript
+interface ActionResult {
+  /** Optional rollback function to undo this action */
+  rollback?: RollbackFunction
+  /** Any data that was mutated/created that can be tracked for assertions */
+  mutations?: Record<string, any>
+}
+```
+
+### FieldTriggerExecutionResult (Enhanced)
```
agritheory: Enhanced from what? This is a new feature and an obvious LLM artifact.

stonecrop/docs/TRANSACTIONAL_TRIGGERS.md
```diff
@@ -0,0 +1,245 @@
...
+  return {
+    mutations: { userId, userCreated: true },
+    rollback: () => {
+      database.delete(userId)  // Undo the user creation
+    }
+  }
+})
+```
+
+### 2. **Mutation Tracking**
```
agritheory: I can see how this would be useful but it's an over specified design if you don't need idempotency, which I would argue we don't and can't support anyway.

agritheory (changes requested):

stonecrop/docs/TRANSACTIONAL_TRIGGERS.md
Alchez: Oops, removed it now :sweat_smile:

stonecrop/docs/FIELD_TRIGGERS.md
```diff
@@ -0,0 +1,894 @@
...
+// This will trigger the 'validateEmailPrimary' action
+store.set('Contact.contact-1.emailAddress.0.is_primary', true)
+
+// This will trigger the 'updateTimestamp' action
+store.set('Contact.contact-1.status', 'completed')
+```
+
+## Action Function Interface
+
+Action functions are simple, focused functions that receive a `FieldChangeContext` object:
```
agritheory, line 117: Worth noting here that there are separate schema driven mask and format functions?

stonecrop/docs/FIELD_TRIGGERS.md
```diff
@@ -0,0 +1,894 @@
...
+  actionResults: ActionExecutionResult[]  // Results for each action
+  totalExecutionTime: number          // Total time for all actions
+  allSucceeded: boolean               // Did all actions succeed?
+  stoppedOnError: boolean             // Did execution stop due to error?
+}
+```
+
+## Pattern Matching
+
+Field trigger patterns support powerful matching capabilities:
```
agritheory: Powerful?

stonecrop/docs/FIELD_TRIGGERS.md
```diff
@@ -0,0 +1,894 @@
...
+})
+```
+
+### 4. Timeouts
+
+Actions have configurable timeouts to prevent hanging:
+
+```typescript
+const engine = new FieldTriggerEngine({
+  defaultTimeout: 5000  // 5 seconds
```
agritheory, line 367: Nice feature.

stonecrop/docs/FIELD_TRIGGERS.md
```diff
@@ -0,0 +1,894 @@
...
+    'checkDuplicates',    // More validation
+    'sendWelcomeEmail',   // Side effects after validation
+    'updateTimestamp'     // Final housekeeping
+  ]
+})
+```
+
+### 4. Don't Assume State
+
+Each action receives only the field change context. Don't assume previous actions succeeded:
```
agritheory: I'm not sure that's great advice. How do you feel about it expressed like this?

> Each action receives only the field change context. While it is possible that a previous function did not succeed and used a `catch` to swallow the error, this is generally considered an antipattern. It is likely that previous actions mutate the document, but not necessarily the state of the individual field.

agritheory (changes requested):
This came out great. Tests look good, though it's possible that we're introducing some bloat but making a full copy of the doc per-field, per-action. I'd like to defer this to a performance optimization ticket that would include measuring memory usage.

stonecrop/docs/FIELD_TRIGGERS.md
Alchez: I would assume that's in lieu of making the docs feel more comprehensive :sweat_smile:

stonecrop/docs/FIELD_TRIGGERS.md
Alchez: I think that's more reliable documentation, yeah. Ultimately the sequence of actions is defined by the user, so this should be good to know for them.

stonecrop/docs/FIELD_TRIGGERS.md
Alchez: Added a note at the end of the section

Alchez:
@agritheory

> Tests look good, though it's possible that we're introducing some bloat but making a full copy of the doc per-field, per-action. I'd like to defer this to a performance optimization ticket that would include measuring memory usage.

The snapshot is only generated once before the actions run, so it is per-field but not per-action. I've created a [ticket](https://github.com/agritheory/stonecrop/issues/309) for future optimization though.

stonecrop/docs/FIELD_TRIGGERS.md
agritheory: You're absolutely right

PR #316, opened 2025-11-13
[nuxt] add Stonecrop playground to Nuxt

### LINKED ISSUE #314: [nuxt] Playground / User Management story / example
DocTypes (editable table view)
DocType (editable form view)
DocBuilder (custom view): /builder/{doctype}

---

Users (table)
User (form)
HasRole (table inside User)
Custom Component for [effective permissions](https://github.com/agritheory/orpin/blob/main/extensions/orpin/orpin--1.0.sql#L712C28-L712C58)

Roles (table)
Role (form)

RoleProfiles (table)
RoleProfile (form)

AbilityRule (form)

---

Docbuilder loads DocType and Ability Rules and linked State Machines

Alchez:
Closes #314


agritheory:
Initial feedback is that this should use the SheetNav and ActionSet components and remove the top bar.
In terms of styling the custom landing page, I think we should go in one of two directions: BEAM/AForm neutral grays, or something that is of similar style to other Nuxt projects

aform/src/components/AForm.vue
```diff
@@ -27,6 +27,20 @@ const { modelValue, data, readonly } = defineProps<{
...
+		formData.value = data
+		// Sync data values into schema
+		modelValue.forEach(field => {
+			if (field.fieldname && data[field.fieldname] !== undefined) {
+				// eslint-disable-next-line vue/no-mutating-props
+				field.value = data[field.fieldname]
+			}
+		})
+	}
+})
```
Alchez, line 42: @agritheory Not having this was causing forms loaded in the playground to not display the HST data. I'm not sure if there's a different, better kind of fix that AForm needs for reactivity, but we can revisit this logic.

PR #320, opened 2026-01-04
[rockfoil] port rockfoil projects

agritheory:
This ports several server-side projects that are designed for the Stonecrop GraphQL stack into the Stonecrop project.


Alchez:
@agritheory I'm leaving this PR in draft since this should ideally go after we've figured out the expired NPM token for release.

agritheory:
Now included in the graphql middleware PR.

PR #334, opened 2026-01-27
[grafserv] add support for postgraphile

Alchez:
Required for https://github.com/agritheory/fab/pull/121

### DIFF
--- a/.github/workflows/nuxt-grafserv.yml
+++ b/.github/workflows/nuxt-grafserv.yml
@@ -1,0 +1,48 @@
+name: Tests
+
+on:
+  push:
+    branches:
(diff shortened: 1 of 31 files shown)


agritheory (changes requested):
Small change, but we probably want to provide an example (and test case) where both Postgraphile and local resolver are used, it's fine if it's a toy example. We need to make sure it's supported.

Alchez:
@agritheory I'm not sure if we can do proper postgraphile examples/tests without an active Postgres database connection. Should we leave the tests for the basic schema/resolvers setup and use FAB as the test for the postgraphile piece of this?

PR #335, opened 2026-01-27
[aform] ADate range Component

### LINKED ISSUE #304: [aform] Datetime component
- [ ] Extend ADate to include a time component, optionally configurable with seconds

crabinak:
PR for the ADate range selection and time components (ADateSelection and ADateTime).


agritheory:
@crabinak Let's add more examples to the input-with-picker story and update the existing one to use the custom picker. 
image

Alchez:
Update: ran into a bunch of compatibility issues when using `<input type=number>`. Basically all of those are documented [here](https://stackoverflow.blog/2022/12/26/why-the-number-input-is-the-worst-input/). I've changed the input type to be text instead, which will also allow padding numbers (0 -> 00).

agritheory:
@IshwaryaM1030 Can you run `rush change` to describe the changes? Please pull down the latest commits first

agritheory (approved):

PR #337, opened 2026-01-30
Fix control row style

crabinak:
I made some fixes to the control row. 

- I was able to fix the spacing issue by just setting the actual cell width within the story (it was set to 8ch, which was too large). I also removed the width style from the index cells, as I believe that was also contributing to large cell width.
- I updated the icons in the action menu for "insert above" and "insert below" to be a bit less heavy by giving them and outline instead of a fill.
- I fixed the bottom border bleed issue on the action menu buttons by moving the outline effect on focus to the cell itself rather than the button. They have the same dimensions, but the outline just seems to behave better on the container rather than on buttons.


agritheory:
@crabinak Still seeing some weird spacing here

image

PR #350, opened 2026-02-18
[graphql_middleware] [nuxt-grafserv] improve graphql middleware and grafserv setup logic

Alchez:
Ran into these issues when setting up the schemas in FAB


agritheory (approved):
Looks good!

PR #356, opened 2026-03-04
[atable] add row-click and row-open event emitters

Alchez:
This came up because FAB needed to handle row-click events, and it was using workarounds using a container `div` element.

---

There's two new stories in the ATable section for clickable and open-able table rows.

image

image


atable/src/icons/stonecrop-ui-icon-open.svg
Alchez, line 1: @agritheory this is something that Claude generated based on the other icons, but we should have a proper design for it

agritheory:
@crabinak Let's add an "Internal link" and "external link" icon to cover these cases (right-pointing arrow and top-right-point arrow)

agritheory:
@Alchez Rename "Open row" to "open record" to disambiguate from "expand row"

PR #364, opened 2026-03-12
[project] remove auto-mask, fix dropdown mismatch, remove sideEffects

Alchez:
**Problem:** All `createdAt`/`modifiedAt` (`Datetime` fieldtype) and `startDate`/`endDate` (`Date` fieldtype) fields show garbled values when a record is loaded from the database. Examples: ISO string `2026-01-01` renders as `20/26/0101`; `2026-03-12T06:11:17.468315+00:00` renders as `2026/03/12 T0:6:`.

**Cause:** The `v-mask` directive in `ATextInput` reads `schema.fieldtype` and auto-applies named masks (`Date` → `##/##/####`, `Datetime` → `####/##/## ##:##`) to the `<input>` element. The mask is applied reactively on value binding, not just on user keystroke events — so existing ISO strings from the database are treated as raw digit input and transformed character-by-character through the mask pattern.

**Solution:** Auto-masking is removed and masks should be explicitly set either using the `mask` property in a doctype schema's field definition or the `v-mask` directive inside a Vue component

### DIFF
--- a/aform/src/components/form/ADropdown.vue
+++ b/aform/src/components/form/ADropdown.vue
@@ -36,30 +36,33 @@
 
 <script setup lang="ts">
 import { vOnClickOutside } from '@vueuse/components'
-import { reactive } from 'vue'
+import { reactive, ref } from 'vue'
 
 import type { ComponentProps } from '../../types'
 
 const {
 	label,
-	items = [],
+	options = [],
 	isAsync = false,
 	filterFunction = undefined,
 	mode,
 } = defineProps<
 	ComponentProps & {
-		items?: string[]
+		options?: string[]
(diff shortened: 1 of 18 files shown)


agritheory:
Disagree with this strategy. These values should be cast into JS date format instead.

Alchez:
@agritheory Do you mean the date value from the db should be cast to the mask format? I think that could work but it may still struggle with some formats (non-automask formats like `##-##-####` or datetime strings). Also we'd have to cast the value back before sending it to the db maybe.

Alternatively maybe a sort of global config we can pass to Stonecrop that auto-sets these global masks? And individual fields can override them using the `mask` schema property?

PR #366, opened 2026-03-16
[nuxt] [stonecrop] improve Stonecrop-Nuxt setup with GraphQL

Alchez:
`@stonecrop/schema`:
- Added `DoctypeContext` interface (renamed from `RouteContext`)
- Added `DoctypeRef` interface - minimal type for record operations
- Added `DataClient` interface with `getMeta`, `getRecord`, `getRecords`, `runAction` methods

`@stonecrop/stonecrop`:
- Added `@stonecrop/schema` as dependency
- Updated `StonecropOptions` to accept optional client
- Added `setClient()`/`getClient()` methods
- Added `dispatchAction()` method for server-side action dispatch
- Updated `getRecord`/`getRecords` to use client, throw error if not configured
- Removed REST fallbacks
- Updated `DoctypeMeta` class with name getter for `DoctypeRef` compatibility
- Updated all tests to use mock `DataClient`

`@stonecrop/nuxt`:
- Updated `useStonecropRegistry()` with `setClient()`/`getClient()`/`dispatchAction()` methods
- Updated documentation with architecture overview and usage examples


graphql_middleware/src/plugin/postgraphile.ts
```diff
@@ -87,9 +87,21 @@ export const createStonecropPlugin = (options: StonecropPluginOptions): Graphile
...
+					align: String
 					required: Boolean
+					readOnly: Boolean
+					edit: Boolean
+					hidden: Boolean
+					default: JSON
 					options: JSON
+					mask: String
+					precision: Int
+					scale: Int
```
agritheory, line 102: What do we use the `scale` property for?

graphql_middleware/src/plugin/postgraphile.ts
```diff
@@ -87,9 +87,21 @@ export const createStonecropPlugin = (options: StonecropPluginOptions): Graphile
...
+					mode: String
```
agritheory, line 103: `mode` as a property name should probably be improved for clarity... I think this was `viewMode` which is just as bad.

graphql_middleware/src/plugin/postgraphile.ts
Alchez: `precision` and `scale` together describe a decimal value in the database, but I don't think it's being used directly anywhere right now. I think this could be used when creating schemas from a db.

graphql_middleware/src/plugin/postgraphile.ts
Alchez: We could use something like `renderMode` but I don't know if that's too close to `viewMode`. Maybe `editState`?

More generally, I'm also wondering if this should become dynamic later and accept a function, for different use-cases? Would/should that change how we describe this config during runtime?

agritheory (approved):

PR #376, opened 2026-03-24
[stonecrop] move framework-only interfaces into class

Alchez:
### Observed Behaviour

When calling `useStonecrop({ doctype, recordId })` inside an async event handler (e.g., a button click handler), the following errors occur:

```
[Vue warn]: inject() can only be used inside setup() or functional components.
[Vue warn]: onMounted is called when there is no active component instance.
[Vue warn]: provide() can only be used inside setup().
Error: HST store not initialized
```

### Root Cause

1. `useStonecrop()` is a Vue composable that uses `inject()`, `onMounted()`, and `provide()` internally
2. Vue composition functions must be called during the component's `setup()` phase, not inside async callbacks
3. The `collectRecordPayload()` function is defined as a closure inside `useStonecrop()`, capturing `hstStore.value` and `stonecrop.value` from the composable's scope
4. There is no alternative API to access this functionality without the Vue composition lifecycle

**Framework code location**: `stonecrop/stonecrop/src/composables/stonecrop.ts:515-568`

```typescript
const collectRecordPayload = (doctype: Doctype, recordId: string): Record<string, any> => {
  if (!hstStore.value || !stonecrop.value) {
    throw new Error('HST store not initialized')  // <-- fails when called outside setup
  }
  // ...
}
```

### Workaround Used

Created a standalone `collectRecordPayload()` function in `app/composables/useDoctypes.ts` that:
- Takes `Stonecrop` instance as a parameter (obtained from `useStonecrop()` called during setup)
- Accesses HST via `stonecrop.getStore()` directly
- Uses `stonecrop.registry.resolveSchema()` directly

```typescript
export function collectRecordPayload(
  stonecrop: Stonecrop,
  doctypeConfig: DoctypeConfig,
  recordId: string,
): Record<string, unknown> {
  const hstStore = stonecrop.getStore()
  const registry = stonecrop.registry
  // ... implementation
}
```

Also duplicated `collectNestedData()` helper since it's not exported from the framework.

### Solution Selected

Add method to `Stonecrop` class**

```typescript
class Stonecrop {
  collectRecordPayload(doctype: Doctype | DoctypeConfig, recordId: string): Record<string, any> {
    const hstStore = this.getStore()
    const schemaArray = typeof doctype === 'string' ? [] : (doctype.fields || [])
    // ... existing implementation
  }
}
```

---

I've also moved a bunch of other composable methods that belong in the Stonecrop class instead.


agritheory:
@Alchez Seems like a moot point if `useStonecrop` points to the Stonecrop singleton, which is instantiated on first load. The singleton pattern might be isolated to the registry but there's no reason to have multiple stores or stonecrop schemas when the HST bridges whatever gaps those might be.

PR #378, opened 2026-03-30
[nuxt] nuxt fullstack example

Alchez:


nuxt/fullstack/app/app.vue
```diff
@@ -1,13 +1,12 @@
 <script setup lang="ts">
 // Fullstack playground demonstrating @stonecrop/nuxt + nuxt-grafserv
-import SheetNav from '../../desktop/src/components/SheetNav.vue'
-import ActionSet from '../../desktop/src/components/ActionSet.vue'
-import type { ActionElements } from '../../desktop/src/types'
+import SheetNav from '../../../desktop/src/components/SheetNav.vue'
```
agritheory, line 3: Shouldn't this import from workspaces instead of directly

agritheory (changes requested):
Small change I think

Issue #360, opened 2026-03-09
[desktop] Update design of ActionSet


agritheory:
The current action set design is not working correctly and needs refinement. A possible location for the action set buttons is the top-right corner, potentially just outside the main content area, as it is a spot for interaction rather than reading. 

I think there are two possible designs here, the first as an improvement of the ~4-5 line height fly-in and a second that is a vertically stacked menu of the available actions. Both of these should be collapsible and use the same rotated multiplication symbol that is used in the sheet nav context

crabinak:
@agritheory Here is some starting design on this. I have the menu as a small box when closed and then expands out using that +/x when toggled. I aligned the top of it with the "modules" section in the main area. Some action items have submenus that can open as well, using the lighter, non-bold gray for hierarchy (submenus would push content down in this situation)

Menu Open by default:
<img width="1415" height="873" alt="Image" src="https://github.com/user-attachments/assets/dc19c018-4e95-4e24-8c00-a0adb3a4639b" />

Menu Closed:

<img width="1415" height="873" alt="Image" src="https://github.com/user-attachments/assets/6b7e6e85-7011-4498-9250-559f587f48ce" />

agritheory:
@crabinak Can you push up these changes? It looks great. The next step is to integrate with the framework, which @Alchez will pick up

PR #390, opened 2026-04-23
added updated styling to action set side menu in stonecrop playground

crabinak:
This makes style changes to the action set style menu. Issue #360 

There is a small bug where if the sub-menu item is open and the user clicks into an action like "new role" the sub-menu will be closed, but the little X will still be in the open position. I was not sure where the best place was to call a manual closing of all sub-menus, or if there is a way to override that behavior so all open sub-menus remain open.


agritheory:
@Alchez Please assist

PR #416, opened 2026-05-26
[schema] Update conversion tool to treat foreign key as Link field

### LINKED ISSUE #410: [schema] Update conversion tool to treat foreign key as Link field
Make field a link fieldtype when it GraphQL (field) type is ID and there is a GraphQL type of the same name. 

```gql
recipeTasks{
  recipe: Id
}
{  __introspection
  RecipeTask
}
```

IshwaryaM1030:
Closes #410


schema/src/converter/heuristics.ts
```diff
@@ -242,6 +241,16 @@ export function classifyFieldType(
 			return base
 		}
 
+		if (namedType.name === 'ID') {
+			const candidateTypeName = toPascalCase(fieldName)
+			if (entityTypes.has(candidateTypeName)) {
+				base.component = 'ALink'
```
Alchez, line 249: `ALink` doesn't exist as a component right now, only `AFormLink`. And we'll also add in `ATableLink` soon enough. Should this component be decided dynamically?

CC @agritheory

Alchez (dismissed):

schema/src/converter/heuristics.ts
agritheory: Right now, it's not super important, the future where there's user-provided link components isn't here yet

Alchez: Fair enough, we can let this one go ahead for now then.

Alchez:
@agritheory @IshwaryaM1030 this one should ideally go in after #415 since there's a PK mismatch that's blocking FAB from upgrading to 0.13

PR #422, opened 2026-06-03
[project] rework schema types in stonecrop

Alchez:
## Rationale
  
The schema package previously had two independently-defined field type systems that had drifted apart:
- `FieldMeta` (used for authoring)
- `SchemaTypes` (used for rendering)

These overlapped in most properties but were defined separately, causing casts and mismatches throughout the codebase. This PR unifies the type architecture with a clean two-stage model: one type for what authors write, and one type for what the renderer consumes.

### Type system

- `FieldMeta` is replaced by `DoctypeField` — a discriminated union of three structural variants: `ValueField` (a scalar or link field), `FieldsetField` (a group of fields), and `TableField` (an inline table with column definitions)
- `SchemaTypes` is replaced by `ResolvedField` — the output of `resolveSchema`(), which expands link relationships and normalises fieldsets and tables into a form `AForm` can render directly
- `FormMode` is renamed to `InteractionMode` and moved to `@stonecrop/schema` so both packages share the same definition
- `Mutable`Doctype is removed; `Doctype.fromObject()` is now the single canonical way to construct a runtime doctype from a plain object or API response

### Authored schema files

- Doctype JSON files no longer require a `kind` property on each field — the schema system infers it automatically from structure (schema → fieldset, columns → table, everything else → scalar). This keeps authored schemas minimal: authors write only the properties that describe what the field is.

### What stays the same

- TypeScript code that constructs `DoctypeField` objects directly (e.g. in test fixtures or programmatic schema builders) still uses `kind: 'field'` — the TypeScript interfaces still require it. Only authored JSON files can omit it.
- `fieldtype` remains on `ValueField` — it carries semantic information used for SQL type mapping, default value initialisation, filter control selection, and link field detection that cannot be expressed through component alone.
- All existing tests pass; the changes are behaviorally equivalent from `AForm`'s perspective.


docs/reference/aform.md
```diff
@@ -207,117 +215,157 @@ export interface AFormLinkValue {
...
-### ComponentProps
+**Properties:**
 
-Defined props for AForm components
+| Property | Type | Description |
+|----------|------|-------------|
+| collapsible? | `boolean` | Whether the fieldset can be collapsed |
+| component? | `string` | Component to render; defaults to `'AFieldset'` |
+| fieldname | `string` | Field identifier |
+| kind | `'fieldset'` | Discriminator |
```
agritheory, line 243: There are already a lot of abstract field names here. `kind` doesn't communicate anything by its name alone, which makes it a bad name. It's as bad as a 'utils' folder. Literally "discriminator" is a better name, but it doesn't explain what it does either.

Alchez: Hmm that's fair. It's closer to field type but that's already taken. Maybe field shape or something similar?

agritheory (approved):

PR #433, opened 2026-06-18
[docbuilder] rework docbuilder tool

Alchez:
image


Alchez:
@agritheory some questions that need an answer for the docbuilder (in no particular order):

- Currently, handlers are added to the registry via server plugins/scripts (ex: `issue:save`, `activate_user`).
  - Do we want to support these in addition to client scripts written (via a JS code-editor field) in the docbuilder itself?
  - How and when should client scripts be added to the registry? Do we add an API that's called automatically on save?
- Should docbuilder be its own package, and/or should we bundle it with the Stonecrop-Nuxt module?
  - Own package is great if we want to support non-Nuxt apps using Stonecrop, but it does mean an explicit install and a second running process (even if invisible to the user).
  - Bundled into Nuxt is great cause we can use the Nuxt DevTools to guide devs to it. No separate install, no separate process.
- Should the docbuilder allow changing fieldtypes and should it validate incompatible changes (Int -> Code Editor, etc.)?
- Should the docbuilder allow renaming and deleting doctypes? That could mean cascading those changes across multiple JSONs via link fields possibly.
- Since the docbuilder would write to files, should we consider undo/redo functionality in the node-editor?

PR #420, opened 2026-06-02
[aform] long text /text box input type

IshwaryaM1030:
Closes #417

PR #434, opened 2026-06-29
Aform longtext styling

crabinak:
Updated styling for the AFormTextboxInput, fixed font issue with inputs not using Arimo (it was defaulting to browser default input styling), and made updates to the checkbox component so it now sizes correctly (label size +20px padding, instead of these massively wide checkbox containers). The checkbox field wasn't correctly styled when disabled, so I fixed that as well. 

I made the AFormTextboxInput a full-width form element, since it spans more than one line. This helps prevent weird layouts when a single-line input is next to a multiline box, as well as giving the user more room to enter information.

I also added a mixed-form-element schema to the stories to showcase the textarea element in the middle of the form with other inputs around it.

### DIFF
--- a/aform/src/components/AForm.vue
+++ b/aform/src/components/AForm.vue
@@ -151,6 +151,7 @@ const childModels = computed(() => childModelsCache.value)
 /* global styles for aform */
 .aform input {
 	font-family: var(--sc-font-family);
+	border: none;
 }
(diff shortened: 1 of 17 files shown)


crabinak:
This is for issue #420. Here is a screenshot of the textarea in a form.
textarea-styling

PR #435, opened 2026-07-06
[AForm] Quantity Component

IshwaryaM1030:
If this pull request is related to a specific package, please append the subject with `[package_name]`, for example `[atable]`. If it makes changes across multiple packages, please append `[project]` to the subject of the pull request.

### DIFF
--- a/aform/api.md
+++ b/aform/api.md
@@ -108,6 +108,14 @@ Vue component exported from @stonecrop/aform.
 import { ANumericInput } from '@stonecrop/aform'
 ```
 
+### AQuantityInput
+
+Vue component exported from @stonecrop/aform.
+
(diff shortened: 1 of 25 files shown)


IshwaryaM1030:
@agritheory @Alchez For your reference
[Screencast from 2026-07-10 14-26-07.webm](https://github.com/user-attachments/assets/4a5645df-567a-483c-b1a4-da0139cda809)

agritheory:
@Alchez How do you feel about using the options key with an object? Should be OK, right? I think we would also want to support a functions here?

Alchez:
@agritheory I'm okay with object-based options. Should be easy to configure (w/ functions) that way.

PR #439, opened 2026-07-23
[aform] Currency Component

### LINKED ISSUE #63: [AForm][ATable]  Currency component
Amount (numeric)
Currency (foreign key/ dropdown)
Base amount (Numeric read-only, computed)
Base currency (Foreign Key read only)
Exchange Rate (read only)

IshwaryaM1030:
Closes #63


Alchez:
@IshwaryaM1030 @agritheory Should this component allow changing both the source and target currencies?

Issue #446, opened 2026-08-15
Support declarative `route` key in doctype schema for custom URL routing


agritheory:
Stonecrop resolves doctypes from URLs using a 1:1 mapping between URL segment and doctype slug. This creates limitations when:

1. **List vs record views need different schemas** — e.g., a doctype list should show 6 curated columns, while the form needs 20+ fields. Today you must either use separate URLs (`/sales-order-list` vs `/sales-order`) or accept both views using the same schema.

2. **URL should differ from internal identity** — e.g., doctype `"SalesOrderList"` with slug `sales-order-list` should handle URL `/sales-order` for the list view.

### Current workarounds

- Use `hidden: true` on fields (hides from *both* list and form)
- Custom `routeStrategy` in host apps (duplicates logic per project)
- Manual route-to-doctype aliasing in `registerMeta` callbacks

### Proposed solution

Add an optional `route` key to `DoctypeMeta`:

```json
{
  "name": "SalesOrderList",
  "slug": "sales-order-list", 
  "route": "/sales-order",
  "fields": [...]
}
```

### Route matching rules

- Explicit `route` values checked before slug-based fallback
- Vue Router-style patterns: `/sales-order` (list), `/sales-order/:id` (record)
- Doctypes without `route` fall back to current `/{slug}` behavior

### Example

| Doctype | route | Matches |
|---------|-------|---------|
| `sales-order-list` | `/sales-order` | List view at `/sales-order` |
| `sales-order` | (none) | Record view at `/sales-order/:id` via slug fallback |

### Touch points

1. `@stonecrop/schema` — Add `route` to `DoctypeMeta` Zod schema
2. `@stonecrop/stonecrop` — Add `route` to `DoctypeConfig` type  
3. `@stonecrop/nuxt` module — Build route map from `route || slug`, index doctypes by route for resolution
4. `useDoctypes` templates — Build reverse map (route → doctype)
5. `registerMeta` callbacks — Resolve by route pattern, not just slug

### Benefits

- **Declarative**: routing co-located with doctype definition
- **Backwards compatible**: existing doctypes work unchanged
- **Pattern exists**: playground plugins already use dual-key registry pattern (URL key for Desktop lookup, computed slug for HST paths)

PR #447, opened 2026-08-15
[nuxt] add routing key for explicit doctype routes

agritheory:
#446


Alchez:
@agritheory I had Claude review the ticket and this PR, and it found that there's some inconsistencies in how the routes are being read:

```
  doctype.json
  ┌──────────────────┐        ┌─► resolveRouteView ──────────► records | record | notFound
  │ name: SalesOrder │──┐     │     rule: name ends "-list"? + how many segments?
  │ route: /orders   │  │     │
  └──────────────────┘  ├─────┼─► resolveDoctypeSlugFromSegments ──► slug
                        │     │     rule: 2 segments? then ignore `route` completely
  URL /orders/SO-1 ─────┘     │
                              └─► resolvePublicUrlSegment ────► path
                                    rule: strip "-list" off any slug, route or not

              three readers · three private rules · nothing reconciles them
```

I've pushed some changes that adds a new `view` key in the doctype JSON that helps define the type of view, so the above changes to:

```
  doctype.json
  ┌─────────────────────┐
  │ route: /orders      │──┐     ┌──────────────────┐
  │ view:  list | form  │  ├────►│   route table    │  built once, whole-app,
  │        | singleton  │  │     │  path → { roles }│  ambiguity throws at boot
  └─────────────────────┘  │     └────────┬─────────┘
                           │              │
  every other doctype ─────┘              ├─► resolve(segments) → { view, slug, recordId }
                                          │
  URL /orders/SO-1 ───────────────────────┘─► pathFor(slug) → path

              one table · one resolve
```

Let me know if those look good, or if we need to revert or re-tune.

PR #454, opened 2026-08-18
[atable][desktop] Pagination

### LINKED ISSUE #344: [graphql-client][graphql-middleware] Support pagination

agritheory:
This PR adds pagination for Desktop: 
- Stonecrop’s `getRecords` now treats `offset === 0` as a replace (clear then add) and `offset > 0` as append, and exposes richer `pageInfo` (`hasMore`, `offset`, `limit`, optional `count`). 
- ATable gets a new `useTablePagination` composable and default footer that paginates *after* filter/sort—pagination is a window over held rows, not another filter—with optional in-memory `pageSize` and server “Load more” via a `getRecords` callback typed from `@stonecrop/schema` so ATable never imports Stonecrop. 
- AForm passes that callback through to list tables; Desktop stops doing its own list fetch and truncation banner and wires `stonecrop.getRecords` instead, so the table owns paging UX while the store owns the record list.
 
Tests cover store replace/append behavior, composable windowing and fetch, ATable footer integration, and Desktop router “Load more” flows. Deliberately out of scope for this PR: `includeTotal` / “showing N of M”, numbered page controls, and route-based page state.


agritheory:
@Alchez Can you review this and see if it aligns with what you had in mind?

PR #456, opened 2026-08-20
[aform] segmented control

agritheory:
image


aform/src/components/base/ExpandButton.vue
Alchez, line 1: @agritheory this isn't being used anywhere, should this go through anyway?

PR #465, opened 2026-09-15
[project] various CSS style cleanup/ modularization

agritheory:
This branch separates product defaults from specimen branding: the token floor in `themes/default/default.css` now uses gray (`--sc-gray-80`) for both `--sc-color-primary` and `--sc-color-brand` instead of the old teal and navy, so real Stonecrop apps default to a neutral gray-on-gray palette while alert colors stay unchanged. The documentation site and Nuxt example apps (`documentation`, `playground`, `fullstack`) share a new unlayered host override at `nuxt/example-host.css` that sets those same primitives to the legacy sage green `#5a8a5a` (the color that had lived only in orphaned SVG assets, not in CSS tokens), wired in via each app’s `nuxt.config.ts` and documented in `content/reference/themes.md`; the fullstack app’s old workaround remapping `--sc-primary-color` to `--sc-brand-color` was removed because primary and brand are aligned again.


Alchez:
@agritheory I've made some changes to resolve the test failures, but they slightly change the styles. Can you check these out locally?

